PR #9, opened 2026-04-18
feat(pipeline): reset CLI — stage, source, and full reset + audit log

parametrization:
## Summary
Implements [noorinalabs/noorinalabs-main#108](https://github.com/noorinalabs/noorinalabs-main/issues/108) — the CLI half. Three reset scopes wired into `isnad-ingest reset`, each emitting an audit entry via the existing `src/pipeline/audit.py` module.

- `isnad-ingest reset --stage {raw|dedup|enriched|normalized|staged}` — wipes one B2 prefix + resets that stage's Kafka consumer-group offsets so the stage can be re-run
- `isnad-ingest reset --source {source-id}` — wipes every stage prefix for one data source (re-acquire from scratch)
- `isnad-ingest reset --all` — wipes every pipeline B2 prefix, clears every pipeline Kafka topic's records (topic *definitions* preserved), truncates Neo4j hadith graph labels, truncates PG hadith metadata tables. **Never** touches users / roles / user_roles / sessions / subscriptions.

Supporting flags: `--yes` (skip the interactive `OBLITERATE` confirmation for `--all`), `--dry-run` (validate scope without touching anything).

## Architecture
- `src/pipeline/reset.py` — pure coordination layer. Adapters injected via `Protocol`s (`KafkaAdmin`, `Neo4jResetter`, `PgResetter`, `_ObjectStoreProto`) so the tests drive the full flow with in-memory fakes.
- `src/pipeline/reset_adapters.py` — concrete `kafka-python` / `neo4j` / `psycopg` adapters. `HADITH_METADATA_TABLES` and `HADITH_NODE_LABELS` are hardcoded allow-lists so user tables can never leak into a truncate path. A guard test asserts this invariant.
- `src/cli.py` — new `reset` subparser with mutually-exclusive mode flags + confirmation prompt.

## Verification
- 10/10 new unit tests passing (no Kafka / S3 / Neo4j / PG required — all fakes)
- 7/7 existing `test_audit.py` still passing
- `ruff check` + `ruff format --check` clean
- CLI validated: `isnad-ingest reset --help` prints mutually-exclusive groups correctly

## Scope note — CLI-only PR
Issue #108 also calls for "Admin dashboard buttons for each reset level (with confirmation modal)" and references "#109 (Admin dashboard data management)". The referenced #109 issue **does not currently exist in the tracker** (it was renumbered to the CI-merge-block hook). This PR delivers the CLI + audit log half; the admin UI buttons should be a separate PR landing once an admin-dashboard data-management issue is filed or re-filed. Flagging this for Program Director.

## TechDebt attestation
- **Admin dashboard UI buttons deferred** — awaits the referenced (missing) admin-dashboard issue.
- **No HTTP endpoint** — reset is CLI-only. If admin dashboard needs to invoke reset, it will either call a new endpoint (follow-up) or shell-exec the CLI (not recommended).
- **Consumer-group offset reset** uses `OffsetAndMetadata(0)` for every partition — benign idempotent rewind, but does not distinguish between "never consumed" and "partition empty".
- **PG COUNT + TRUNCATE race** — audit count is read just before truncate, so writes arriving in between would be under-counted. Acceptable for admin-initiated resets where the caller stops producers first; noted for future hardening (table lock or atomic `TRUNCATE ... RETURNING` equivalent).

## Disabled CI jobs
None.

## Review
Review requested (2-reviewer charter): @Santiago Ferreira (release coordinator — infra touchpoints) and @Nurul Hakim (per team-lead assignment).

---
Part of noorinalabs/noorinalabs-main#108

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/cli.py
+++ b/src/cli.py
@@ -395,6 +395,121 @@ def _cmd_audit(*, last_n: int = 10) -> None:
         print()
 
 
+def _cmd_reset(
+    *,
+    stage: str | None,
+    source: str | None,
+    full: bool,
+    assume_yes: bool,
+    dry_run: bool,
+) -> None:
+    """Execute a pipeline reset. Writes an audit entry for every run."""
+    from pathlib import Path
+
+    from src.config import get_settings
+    from src.pipeline.reset import PipelineResetter, ResetScope
+
+    # Build the scope first — validates inputs before any confirmation.
+    provided = sum(1 for f in (stage, source, full) if f)
+    if provided != 1:
+        print("ERROR: reset requires exactly one of --stage, --source, --all.")
+        sys.exit(2)
+
+    if stage:
+        scope = ResetScope.stage_scope(stage)
+        summary_line = f"STAGE reset: prefix '{stage}/' and consumer offsets"
+    elif source:
+        scope = ResetScope.source_scope(source)
+        summary_line = f"SOURCE reset: all stage prefixes for source '{source}'"
+    else:
+        scope = ResetScope.full_scope()
+        summary_line = (
+            "FULL reset: every pipeline B2 prefix, every pipeline Kafka topic, "
+            "Neo4j hadith graph, PG hadith metadata. "
+            "Users/roles/sessions are PRESERVED."
+        )
+
+    print(f"=== Pipeline Reset ===\n  {summary_line}\n")
+
+    if scope.level == "full" and not assume_yes:
+        confirm = input("Type 'OBLITERATE' to proceed: ").strip()
+        if confirm != "OBLITERATE":
+            print("Aborted.")
+            sys.exit(1)
+
+    if dry_run:
+        print("Dry run — no changes were made. Scope validated.")
+        return
+
+    settings = get_settings()
+    object_store, kafka_admin, neo4j, pg = _build_reset_clients(settings)
+
+    resetter = PipelineResetter(
+        object_store=object_store,
+        kafka_admin=kafka_admin,
+        neo4j=neo4j,
+        pg=pg,
+        data_dir=Path(settings.data_raw_dir).parent,
+    )
+    report, _entry, audit_path = resetter.reset(scope)
+
+    print("\n=== Reset Complete ===")
+    print(f"  Level              : {report.level}")
+    print(f"  S3 prefixes deleted: {len(report.s3_prefixes_deleted)}")
+    print(f"  S3 objects deleted : {report.s3_objects_deleted}")
+    print(f"  Kafka topics reset : {len(report.kafka_topics_reset)}")
+    if report.level == "full":
+        print(f"  Neo4j rows deleted : {report.neo4j_rows_deleted}")
+        print(f"  PG rows deleted    : {report.pg_rows_deleted}")
+    print(f"  Duration (s)       : {report.duration_seconds}")
+    print(f"  Audit entry        : {audit_path}")
+
+
+def _build_reset_clients(settings: object) -> tuple[object, object, object, object]:
+    """Construct the four reset adapters from settings. Isolated for testability."""
+    # Imports kept inside the helper so unit tests that call _cmd_reset
+    # via monkey-patched adapters don't drag in kafka/neo4j/boto3.
+    import os
+
+    import boto3  # type: ignore[import-untyped]
+    import psycopg
+    from kafka.admin import KafkaAdminClient  # type: ignore[import-untyped]
+    from neo4j import GraphDatabase
+
+    from src.pipeline.reset_adapters import (
+        KafkaPythonAdmin,
+        Neo4jHadithResetter,
+        PgHadithMetadataResetter,
+    )
+
+    class _Store:
+        def __init__(self, bucket: str, endpoint_url: str | None) -> None:
+            self.bucket = bucket
+            self._client = boto3.client(
+                "s3", **({"endpoint_url": endpoint_url} if endpoint_url else {})
+            )
+
+        @property
+        def client(self) -> object:
+            return self._client
+
+    bucket = os.environ.get("PIPELINE_BUCKET", "noorinalabs-pipeline")
+    endpoint_url = os.environ.get("S3_ENDPOINT_URL")
+    bootstrap = os.environ["KAFKA_BOOTSTRAP_SERVERS"]
+
+    neo4j_cfg = getattr(settings, "neo4j")
+    pg_cfg = getattr(settings, "postgres")
+
+    return (
+        _Store(bucket=bucket, endpoint_url=endpoint_url),
+        KafkaPythonAdmin(KafkaAdminClient(bootstrap_servers=bootstrap)),
+        Neo4jHadithResetter(
+            GraphDatabase.driver(neo4j_cfg.uri, auth=(neo4j_cfg.user, neo4j_cfg.password))
+        ),
+        PgHadithMetadataResetter(psycopg.connect(pg_cfg.dsn)),
+    )
+
+
 def _cmd_pipeline() -> None:
     """Run the full pipeline: acquire -> parse -> resolve -> load -> enrich."""
     print("=== Full Pipeline Run ===\n")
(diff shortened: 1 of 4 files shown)


parametrization:
Review requested (2-reviewer charter): @Santiago Ferreira (release coordinator — infra touchpoints) and @Nurul Hakim.

parametrization:
RequestOrReplied: Replied
TechDebt: none
Reviewer: Santiago Ferreira

Local verification (release-coord seat):

**Tests** — 17/17 passing on `W.Mwangi/0108-pipeline-reset` @ `2e03d7c`:
- 10/10 new in `tests/test_pipeline/test_reset.py`
- 7/7 existing in `tests/test_pipeline/test_audit.py`
- `ruff check` + `ruff format --check` clean on all four changed files

**`reset --all` safety — verified**
- `HADITH_METADATA_TABLES` and `HADITH_NODE_LABELS` are module-level `tuple[str, ...]` constants in `src/pipeline/reset_adapters.py:16-34`. `PgHadithMetadataResetter.truncate_hadith_metadata` and `Neo4jHadithResetter.truncate_hadith_graph` iterate ONLY those constants — neither method accepts a table/label parameter. There is no code path by which a user table can leak in.
- `test_full_reset_does_not_touch_user_tables_by_contract` asserts `set(HADITH_METADATA_TABLES) & {users, roles, user_roles, sessions, subscriptions} == set()`. Guard is real and load-bearing.
- `OBLITERATE` confirmation in `src/cli.py:434-438` is un-skippable when `--yes` is omitted: `input()` blocks on stdin, mismatch (e.g. typing `no`) hits `sys.exit(1)`. Verified live: `echo 'no' | isnad-ingest reset --all` → "Aborted." exit 1.

**`--dry-run` — verified side-effect-free**
- Path returns at `src/cli.py:440-442` BEFORE `_build_reset_clients(settings)` is invoked. The boto3 / kafka / neo4j / psycopg imports live inside `_build_reset_clients`, so a dry-run doesn't even import them. No S3, Kafka admin, Neo4j, or PG client is constructed.
- Verified live: `OBLITERATE | reset --all --dry-run` prints "Dry run — no changes were made. Scope validated." exit 0, with no infra reachable from this shell.

**Audit emission — every branch covered**
- `PipelineResetter.reset()` (`reset.py:184-210`) is the single entrypoint. All three branches (`_reset_stage`, `_reset_source`, `_reset_full`) return a `ResetReport`, then audit is built + written at lines 203-209 before the tuple returns. Impossible to take a reset path that skips audit emission.
- `test_audit_entry_is_written_per_reset` confirms two stage resets produce two audit files; `test_full_reset_wipes_everything_and_calls_neo4j_and_pg` asserts `audit_path.exists()` after a full reset.

**Minor non-blocking notes**
- `_cmd_reset` lines 413-416: `provided = sum(1 for f in (stage, source, full) if f); if provided != 1` is dead defensive code — argparse's `add_mutually_exclusive_group(required=True)` already enforces exactly-one. Harmless overlap; not blocking.
- TechDebt items in PR body (offset rewind semantics, PG COUNT/TRUNCATE race) are appropriately scoped and acceptable for admin-initiated resets where producers are stopped first. No follow-up issues needed pre-merge.

**Missing #109 — disposition**
Issue #109 referenced in #108 doesn't exist (renumbered to the CI-merge-block hook). Author flagged this for Program Director — agreed. My view: the admin-dashboard UI buttons should be a fresh issue filed after this CLI lands (CLI is the substrate the UI will call, so the ordering is correct). Not a merge blocker for this PR — the CLI + audit half is fully usable standalone. Recommend Program Director files a new "Admin dashboard data-management UI (reset buttons)" issue against `noorinalabs-isnad-graph` (frontend) for a future wave, linking back to the now-merged CLI.

parametrization:
RequestOrReplied: Replied
TechDebt: #15, #16
Reviewer: Nurul Hakim

Observability seat. Read `audit.py`, `reset.py`, `reset_adapters.py`, and the `_cmd_reset` call-site. Notes below — none blocking, all filed as followups so this PR can land.

**Audit log structure** (`reset.py:388-394`, summary into `audit.py:AuditEntry`)
The schema is solid for a v1 SoT — `stage`, `timestamp` (UTC ISO8601), `operator`, `duration_seconds`, `rows_affected`, plus the level-aware summary dict. Three gaps for downstream querying, all filed as #16:
- `dry_run` is not recorded — and dry-runs skip audit entirely. Capturing dry-runs gives us "operator practiced first" telemetry, which is exactly the kind of pre-incident signal SIEM/Grafana want.
- `confirmation_method` (`interactive` vs `yes-flag` vs `not-required`) — operationally distinct (runbook script vs human-at-keyboard) but indistinguishable in audit today.
- `scope.stage` / `scope.source` only inferable by substring-matching `s3_prefixes_deleted`. Field-equality queries ("every reset that touched `sunnah-api`") should not require regex over a JSON-serialized list.

Also flagged in #16 as a bonus consideration: `getpass.getuser()` becomes `root` / service-account on containers and CI runners. Capturing `SUDO_USER` / `GITHUB_ACTOR` / etc. when present makes attribution to a human possible. Split off if you'd rather treat it separately.

**Logging hygiene** (`cli.py:_cmd_reset`)
All output is `print()`. The rest of `cli.py` is the same (consistent with `_cmd_audit` adjacency) so this PR isn't regressive. Long-term, the CLI deserves a stdlib `logging` migration so reset summaries emit at INFO and errors at ERROR with proper streams (currently the error path on line 26 `print("ERROR: ...")` goes to stdout, not stderr — minor but real for log-aggregation pipelines that key on stream). Not blocking this PR; worth a CLI-wide cleanup issue if one doesn't exist.

**PG COUNT + TRUNCATE race** (`reset_adapters.py:583-596`)
Concur with the author's call. For admin-initiated resets where producers are stopped, the under-count window is operationally negligible. The hardened pattern (`BEGIN; LOCK TABLE x IN ACCESS EXCLUSIVE MODE; SELECT count(*); TRUNCATE; COMMIT;` per table) changes the failure mode from "audit slightly under-counts" to "reset blocks on any open transaction" — that's a worse trade today. Document the precondition (stop producers first) in the runbook when the admin-dashboard UI ships and we'd be set.

One small note on the truncate path: `TRUNCATE TABLE x CASCADE` runs per table in a loop, then a single `commit()` at the end. If any table's truncate fails mid-loop, half the tables are wiped + uncommitted, the operator sees an exception, and the next reset attempt finds inconsistent state. Wrapping the loop in an explicit `BEGIN`/`ROLLBACK ON ERROR` would make the operation atomic-or-nothing. Not blocking — file as part of the eventual hardening pass alongside the lock pattern.

**Dry-run output verbosity** (`cli.py:51-53`)
This was the find that bothers me most. Dry-run prints exactly:
```
Dry run — no changes were made. Scope validated.
```
For `--all`, this defeats the safety value entirely — operator can't sanity-check that "20 prefixes, 5 topics, 7 Neo4j labels, 6 PG tables" matches their intent before re-running without `--dry-run`. Filed as #15. Suggested fix: pull scope-summarization into a `scope.describe()` method and have both the dry-run path and the post-execution report share it. Cheap to add now, expensive to add after operators have built habits around vague output.

**Things that look good**
- `HADITH_METADATA_TABLES` and `HADITH_NODE_LABELS` as hardcoded allow-lists with the guard test in `test_full_reset_does_not_touch_user_tables_by_contract` is exactly the right shape — invariant is in code, drift is caught by CI (when CI exists).
- Paginated `_delete_prefix` with `_PagingClient` test coverage — the realistic case where a prefix exceeds one `list_objects_v2` response is exercised.
- `delete_topic_data` preserves topic config (partitions, retention, compaction) — important for not having to re-tune Kafka after every full reset.
- Adapter injection via `Protocol`s makes the test suite trivially fast and hermetic. `_FakeS3Client` / `_FakeKafkaAdmin` / `_FakeNeo4j` / `_FakePg` all in-process — no infra needed to validate the coordination logic. This is the pattern this repo should extend everywhere.
- `OBLITERATE` confirmation gate on `--all` only (not on `--stage` / `--source`) is right-sized friction.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

Approved. 17/17 tests pass locally; user-table guard, OBLITERATE prompt, dry-run side-effect-free path, and per-branch audit emission all verified. Defer admin-dashboard UI to a fresh issue (Program Director attention on missing #109); not a merge blocker.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Nurul Hakim
RequestOrReplied: Approved
TechDebt: #15, #16

Approving — audit structure is good v1 (operator/timestamp/duration/scope-counts captured), kill-switch invariants for user tables are encoded in code with a guard test, and adapter injection lets the suite cover full reset flows without infra. Followups #15 (dry-run scope verbosity) and #16 (audit fields: dry_run / confirmation_method / discrete scope) are non-blocking — file before operators build habits around vague dry-run output.

Issue #10, opened 2026-04-18
feat: port real processor logic into worker shells (dedup, enrich, normalize, graph-load)


parametrization:
## Summary

The worker scaffolds in the `W.Mwangi-0107-kafka-workers` worktree (tracked by main#107) have happy-path Kafka topology but the four processors are placeholder pass-throughs. The real implementations exist elsewhere in this repo and need to be wired in.

## Sources to port

| Worker | Current | Target logic |
|---|---|---|
| `workers/dedup/processor.py` | TODO pass-through | `src/resolve/dedup.py` (FAISS index, variant tiering) |
| `workers/enrich/processor.py` | TODO pass-through | `src/enrich/metrics.py`, `src/enrich/topics.py` |
| `workers/normalize/processor.py` | TODO placeholder | New stage; schema and logic to be drafted |
| `workers/ingest/processor.py` | reads Parquet, no Neo4j write | `src/graph/load_nodes.py`, `src/graph/load_edges.py` via `load_all()` in `src/graph/__init__.py` |

## Scope

- Each worker consumes its input topic, runs its real logic, emits to its output topic (or DLQ on failure)
- `ingest` worker takes a Neo4j driver via DI (creds from env per decision ticket in main)
- Swap the in-memory idempotency checkpoint for a durable store (Redis or Postgres) before prod
- Preserve MERGE-on-ID + explicit SET safety from Phase 4

## Dependencies

- Blocked by: main#107 (worker scaffolds PR'd), noorinalabs-deploy#80 (Kafka), noorinalabs-deploy#79 (B2/MinIO)
- Depends on decisions in noorinalabs-main (Kafka topology, Neo4j creds, B2 policy)

## Acceptance

- All four workers pass their existing unit tests
- Integration test (via main#136) exercises full pipeline locally end-to-end on `sunnah_api` fixture data
- `validate.py` passes against loaded Neo4j after `ingest` completes

PR #18, opened 2026-04-22
feat(ingest-worker): open Neo4j session, MERGE on stable ID, retryable transactions (#13)

### LINKED ISSUE #13: tech-debt: ingest-worker — open Neo4j session and execute MERGE in processor
## Context
PR #8's `workers/ingest/main.py` constructs a Neo4j driver and passes it into the processor, but `workers/ingest/processor.py` does not yet open a session or run any Cypher. The processor is currently a no-op terminal stage.

This is part of the broader processor-port work tracked in #10, but specifically the Neo4j-driver glue (driver lifecycle, session-per-batch, retryable transaction, MERGE on stable IDs from the unified normalize schema) is its own slice — landing it unblocks end-to-end pipeline smoke tests independent of the dedup/enrich/normalize ports.

## Acceptance
- `processor.py` opens a session per batch and MERGEs nodes/edges from the normalized Parquet
- Driver is closed cleanly on worker shutdown (signal handler)
- Integration test against a Neo4j docker-compose service writes and reads back at least one node

parametrization:
## Summary

Closes #13. Turns the ingest-worker processor from a no-op into a real terminal stage that opens a Neo4j session per batch and MERGEs nodes from the normalized Parquet on stable IDs.

- **Session per batch.** Each Kafka message → one session → one retryable write tx via `session.execute_write`. No custom retry loop; the neo4j driver handles `ServiceUnavailable` internally.
- **MERGE on stable IDs from the unified normalize schema** (landing in #10). Each row carries `label` (validated against an allowlist to prevent Cypher injection) and `id`; remaining columns become node properties via explicit `SET n += row.props`. Per-label UNWIND batches.
- **Driver lifecycle** (in `workers/ingest/main.py`):
  - `driver.verify_connectivity()` at startup — fail-fast if Neo4j is unreachable or creds are bad.
  - SIGTERM/SIGINT handler closes the consumer (letting `run_forever` exit naturally) and the driver.
  - All lifecycle stays in `main.py` — `workers/lib/runner.py` is untouched to avoid colliding with #10.

## Error taxonomy

| Exception | Handling |
|---|---|
| `ServiceUnavailable` | Retried inside `execute_write`; if it still fails it propagates → runner DLQ |
| `ClientError` (bad Cypher / constraint violation) | No retry; propagates → runner DLQ with metadata |
| `SessionExpired` | Reopen session once, retry |
| `UnknownSchemaError` (new) | Raised when a row lacks valid `label`/`id`; runner DLQ |
| Anything else | Propagates → runner DLQ via `build_dlq_record` |

## Scope

- ONLY `workers/ingest/` — dedup/enrich/normalize workers are Wanjiku's #10 territory (parallel branch `W.Mwangi/0010-processor-port`).
- `workers/lib/*.py` untouched.

## Test coverage

10 new unit tests in `tests/workers/test_ingest_neo4j.py`:

- `_merge_node_tx` Cypher shape (param binding, SET shape, return count)
- Happy-path merge-by-label (multi-label batch, props stripped of `label`/`id`)
- Empty Parquet → no-op
- No-driver → read-only (unit-suite ergonomics)
- Unknown label → `UnknownSchemaError`, zero Neo4j calls
- Missing `id` → `UnknownSchemaError`
- `SessionExpired` → reopen-once retry succeeds
- `ClientError` → propagates (no silent retry)
- End-to-end `WorkerRunner.handle_one` routes `ClientError` to DLQ with correct `error_class`
- End-to-end `WorkerRunner.handle_one` routes `UnknownSchemaError` to DLQ

All use fake driver/session — no Docker, no live Neo4j.

**Integration test against testcontainers Neo4j is deferred to #136** (noted in issue #13 acceptance as a follow-up). `testcontainers[neo4j]` is already in dev deps, so that issue can add a container-backed round-trip without new deps.

## Test plan

- [x] `uv run pytest tests/workers/test_ingest_neo4j.py tests/workers/test_processors.py tests/workers/test_runner.py` — 20 passed
- [x] `uv run ruff check workers/ingest/ tests/workers/test_ingest_neo4j.py` — clean
- [x] `uv run ruff format --check workers/ingest/ tests/workers/test_ingest_neo4j.py` — clean
- [x] `uv run mypy workers/ingest/` — Success, no issues
- [ ] Integration: testcontainers Neo4j round-trip (deferred to #136)

## Notes

- `uv.lock` updated because `uv sync` resolved new transitives (`boto3` chain) that the existing deps require; no `pyproject.toml` changes.
- `UnknownSchemaError` is a subclass of `ValueError` so existing `except ValueError` handlers continue to work.

Refs #13.

### DIFF
--- a/tests/workers/test_ingest_neo4j.py
+++ b/tests/workers/test_ingest_neo4j.py
@@ -1,0 +1,324 @@
+"""Unit tests for the Neo4j-backed ingest processor (issue #13).
+
+The tests mock the neo4j driver/session so the suite runs without
+Docker or a live database. Integration coverage against a real Neo4j
+(testcontainers) is tracked separately in issue #136.
+"""
+
+from __future__ import annotations
+
+import io
+import json
+from typing import Any
+
+import pyarrow as pa
+import pyarrow.parquet as pq
+import pytest
+
+from tests.workers.conftest import FakeConsumer, FakeProducer
+from workers.ingest.processor import (
+    IngestProcessor,
+    UnknownSchemaError,
+    _merge_node_tx,
+)
+from workers.lib.dlq import DLQ_TOPIC
+from workers.lib.message import PipelineMessage, serialize_message
+from workers.lib.object_store import ObjectStore
+from workers.lib.runner import WorkerRunner, WorkerSettings
+
+# ---------------------------------------------------------------------------
+# Neo4j driver fakes
+# ---------------------------------------------------------------------------
+
+
+class _FakeResult:
+    def __init__(self, record: dict[str, Any] | None) -> None:
+        self._record = record
+
+    def single(self) -> dict[str, Any] | None:
+        return self._record
+
+
+class _FakeTx:
+    def __init__(self, sink: list[tuple[str, dict[str, Any]]]) -> None:
+        self._sink = sink
+
+    def run(self, query: str, **params: Any) -> _FakeResult:
+        self._sink.append((query, params))
+        rows = params.get("rows") or []
+        return _FakeResult({"merged": len(rows)})
+
+
+class _FakeSession:
+    def __init__(
+        self,
+        sink: list[tuple[str, dict[str, Any]]],
+        *,
+        raise_on_execute: Exception | None = None,
+    ) -> None:
+        self._sink = sink
+        self._raise = raise_on_execute
+
+    def __enter__(self) -> _FakeSession:
+        return self
+
+    def __exit__(self, *exc: object) -> None:
+        return None
+
+    def execute_write(self, fn: Any, **kwargs: Any) -> Any:
+        if self._raise is not None:
+            err = self._raise
+            self._raise = None  # one-shot: subsequent calls succeed
+            raise err
+        tx = _FakeTx(self._sink)
+        return fn(tx, **kwargs)
+
+
+class _FakeDriver:
+    def __init__(self, *, session_factories: list[_FakeSession] | None = None) -> None:
+        self.sink: list[tuple[str, dict[str, Any]]] = []
+        self.closed = False
+        self._factories = session_factories or []
+        self._next_factory = 0
+
+    def session(self) -> _FakeSession:
+        if self._factories:
+            if self._next_factory >= len(self._factories):
+                msg = "No more fake sessions configured"
+                raise AssertionError(msg)
+            sess = self._factories[self._next_factory]
+            self._next_factory += 1
+            sess._sink = self.sink  # share recording sink
+            return sess
+        return _FakeSession(self.sink)
+
+    def close(self) -> None:
+        self.closed = True
+
+
+# ---------------------------------------------------------------------------
+# Parquet payload helpers
+# ---------------------------------------------------------------------------
+
+
+def _parquet_bytes(rows: list[dict[str, Any]]) -> bytes:
+    if not rows:
+        table = pa.table({"label": [], "id": []})
+    else:
+        columns: dict[str, list[Any]] = {}
+        for row in rows:
+            for k, v in row.items():
+                columns.setdefault(k, [])
+        for row in rows:
+            for k in columns:
+                columns[k].append(row.get(k))
+        table = pa.table(columns)
+    buf = io.BytesIO()
+    pq.write_table(table, buf)
+    return buf.getvalue()
+
+
+# ---------------------------------------------------------------------------
+# _merge_node_tx — Cypher shape
+# ---------------------------------------------------------------------------
+
+
+def test_merge_node_tx_emits_parameterized_merge() -> None:
+    sink: list[tuple[str, dict[str, Any]]] = []
+    tx = _FakeTx(sink)
+    rows = [
+        {"id": "nar:ibn-abbas", "props": {"name_en": "Ibn Abbas"}},
+        {"id": "nar:abu-hurayra", "props": {"name_en": "Abu Hurayra"}},
+    ]
+
+    count = _merge_node_tx(tx, label="Narrator", rows=rows)
+
+    assert count == 2
+    assert len(sink) == 1
+    query, params = sink[0]
+    assert "MERGE (n:`Narrator` {id: row.id})" in query
+    assert "SET n += row.props" in query
+    assert params["rows"] == rows
+
+
+# ---------------------------------------------------------------------------
+# IngestProcessor — happy path
+# ---------------------------------------------------------------------------
+
+
+def test_ingest_opens_session_and_merges_by_label(
+    object_store: ObjectStore, sample_message: PipelineMessage
+) -> None:
+    payload = _parquet_bytes(
+        [
+            {"label": "Narrator", "id": "nar:a", "name_en": "A"},
+            {"label": "Narrator", "id": "nar:b", "name_en": "B"},
+            {"label": "Hadith", "id": "hdt:1", "matn_ar": "..."},
+        ]
+    )
+    object_store.put_object(sample_message.b2_path, payload)
+    driver = _FakeDriver()
+
+    result = IngestProcessor(object_store, neo4j_driver=driver)(sample_message)
+
+    assert result is None  # terminal stage
+    # One MERGE statement per distinct label → 2 tx.run calls
+    assert len(driver.sink) == 2
+    queries = [q for q, _ in driver.sink]
+    assert any("MERGE (n:`Narrator` {id: row.id})" in q for q in queries)
+    assert any("MERGE (n:`Hadith` {id: row.id})" in q for q in queries)
+    # Narrator batch carries both rows with props stripped of label/id
+    narrator_batch = next(params["rows"] for q, params in driver.sink if "Narrator" in q)
+    assert {r["id"] for r in narrator_batch} == {"nar:a", "nar:b"}
+    for row in narrator_batch:
+        assert "label" not in row["props"]
+        assert "id" not in row["props"]
+
+
+def test_ingest_empty_parquet_is_noop(
+    object_store: ObjectStore, sample_message: PipelineMessage
+) -> None:
+    object_store.put_object(sample_message.b2_path, _parquet_bytes([]))
+    driver = _FakeDriver()
+
+    IngestProcessor(object_store, neo4j_driver=driver)(sample_message)
+
+    assert driver.sink == []
+
+
+def test_ingest_without_driver_is_readonly(
+    object_store: ObjectStore, sample_message: PipelineMessage
+) -> None:
+    """Unit-suite ergonomics: no driver → read payload, skip write."""
+    object_store.put_object(sample_message.b2_path, _parquet_bytes([]))
+
+    # No assertion on Neo4j; absence of exception is the contract.
+    IngestProcessor(object_store, neo4j_driver=None)(sample_message)
+
+
+# ---------------------------------------------------------------------------
+# IngestProcessor — error taxonomy
+# ---------------------------------------------------------------------------
+
+
+def test_ingest_unknown_label_raises_schema_error(
+    object_store: ObjectStore, sample_message: PipelineMessage
+) -> None:
+    payload = _parquet_bytes([{"label": "Evil", "id": "x:1"}])
+    object_store.put_object(sample_message.b2_path, payload)
+    driver = _FakeDriver()
+
+    with pytest.raises(UnknownSchemaError, match="unknown label"):
+        IngestProcessor(object_store, neo4j_driver=driver)(sample_message)
+
+    assert driver.sink == []  # no Neo4j call attempted
+
+
+def test_ingest_missing_id_raises_schema_error(
+    object_store: ObjectStore, sample_message: PipelineMessage
+) -> None:
+    payload = _parquet_bytes([{"label": "Narrator", "id": None}])
+    object_store.put_object(sample_message.b2_path, payload)
+    driver = _FakeDriver()
+
+    with pytest.raises(UnknownSchemaError, match="'id'"):
+        IngestProcessor(object_store, neo4j_driver=driver)(sample_message)
+
+
+def test_ingest_retries_once_on_session_expired(
+    object_store: ObjectStore, sample_message: PipelineMessage
+) -> None:
+    from neo4j.exceptions import SessionExpired
+
+    payload = _parquet_bytes([{"label": "Narrator", "id": "nar:a", "name_en": "A"}])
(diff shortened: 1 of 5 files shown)


parametrization:
Requestor: Weronika Zielinska
Requestee: Arjun Raghavan
RequestOrReplied: Request
TechDebt: none

Arjun — Neo4j domain review, please. Areas to focus on:

1. **MERGE semantics & property update.** `processor.py` uses `MERGE (n:`Label` {id: row.id}) SET n += row.props` per-label UNWIND batch. This matches the pattern in `src/graph/load_nodes.py` but uses `+=` rather than the explicit `SET n.key = row.key,...` Phase 4 form. The tradeoff: `+=` is generic across the unified schema (#10 hasn't fixed the columns yet), and it only touches keys present in `props`. If you'd rather we enumerate per label after #10 lands the schema, happy to file a follow-up.
2. **Retry policy.** We rely on `session.execute_write`'s built-in retry for `ServiceUnavailable`. `SessionExpired` gets one manual reopen-and-retry. `ClientError` is treated as terminal (DLQ). Does that match how isnad-graph runs the existing loaders?
3. **Cypher injection guard.** Label interpolation is protected by the `_ALLOWED_LABELS` frozenset. Any labels you'd add/remove before #10 schema lands?

parametrization:
Requestor: Weronika Zielinska
Requestee: Farhan Malik
RequestOrReplied: Request
TechDebt: none

Farhan — schema-correctness review, please. The key call on your side:

1. **Stable-ID assumption.** The processor reads normalized Parquet rows and expects each row to carry `label` (string) and `id` (string, already-prefixed per the existing loaders — `nar:`, `hdt:`, `col:`, `chn:`, `grd:`, etc). Remaining columns become node properties. Is that the contract #10 will ship, or should I adjust the field names (e.g. `node_type` instead of `label`, or `canonical_id` instead of `id`)?
2. **Allowed label set.** Currently `_ALLOWED_LABELS = {Narrator, Hadith, Collection, Chain, Grading, HistoricalEvent, Location}` — same seven as `src/graph/load_nodes.py`. Does that cover everything the normalize stage will produce?
3. **Edges deferred.** This PR loads nodes only. Edges (`TRANSMITTED_TO`, `NARRATED`, `APPEARS_IN`, `PARALLEL_OF`, `STUDIED_UNDER`, `GRADED_BY`) per `src/graph/load_edges.py` are not ported yet — intentional, tracked as follow-up. Any objection to doing that in a separate issue once #10 is merged?

parametrization:
Requestor: Weronika.Zielinska
Requestee: Arjun.Raghavan
RequestOrReplied: Approved
TechDebt: #19, #20

Approved. Clean MERGE shape — parameterized `id`/props, backticked label after allowlist check, `RETURN count` for observability. Idempotent; SIGTERM mid-tx leaves no partial graph state (MERGE is atomic per-tx; Kafka offset uncommitted → replay on restart, which MERGE absorbs).

Two architectural concerns filed, neither blocks merge:

- **#19 schema drift**: `_ALLOWED_LABELS` is a hardcoded frozenset, decoupled from #10's unified normalize schema. New entity type in #10 → every row DLQs loudly until someone edits processor.py. Correct failure mode, wrong single-source-of-truth.
- **#20 retry compounding**: `execute_write`'s 30s `max_transaction_retry_time` + the manual `SessionExpired` reopen = ~60s unbounded worst case per message. Consumer stalls instead of shedding to DLQ under sustained Neo4j degradation. Also: verify the manual reopen is still needed with neo4j-driver 5.15+ — driver likely handles `SessionExpired` internally now.

ClientError DLQ payload is rich enough — `error_class` preserves the subclass name (`ConstraintError` vs `CypherSyntaxError` vs `Forbidden`), plus message and traceback. Injection guard is sound.

parametrization:
Requestor: Weronika.Zielinska
Requestee: Farhan.Malik
RequestOrReplied: Changes Requested
TechDebt: #22

Changes Requested — schema/stable-ID contract issues:

1. **Phase 4 safety rule violated.** Docstring (L557) claims "explicit SET-per-key (Phase 4 safety rule from `src/graph/load_nodes.py`)" but `_merge_node_tx` does `SET n += row.props` — the exact `SET n += row` anti-pattern the reference file prohibits ("with explicit property SET (no `SET n += row`) for Phase 4 safety"). Either match the reference (one MERGE stanza per label with enumerated `SET n.name_ar = row.name_ar, …`) or explicitly decide to drop the rule and update the docstring. Silent `+=` will overwrite scholar-curated fields on re-ingest.

2. **Schema mismatch with normalize (#10).** `_group_rows_by_label` requires `row.label` and `row.id`, but the current normalize processor produces `HADITH_SCHEMA` rows with neither column — it emits one flat hadith row per input, no label fan-out. This PR's happy path will hit `UnknownSchemaError` on every real message until #10 is redesigned to emit the multi-label `{label, id, props}` shape. Please coordinate the unified normalize schema with Wanjiku before merge and add a docstring contract in `processor.py`.

3. **Stable-ID source undocumented.** `MERGE ({id: row.id})` — `id` origin isn't specified. For MERGE to actually merge (not duplicate) on re-ingest, `id` must be a deterministic hash of source+type+content, generated in normalize. Add a contract comment and a test asserting identical input → identical `row.id`.

4. Label allowlist matches `ontology/services.yaml` — LGTM. Edge MERGE followup filed as #22.

parametrization:
**Deferred pending architectural reconciliation with #21.**

Parking this PR open. Reviewer context (Arjun ✅ with #19 + #20 filed; Farhan ⚠️ Changes Requested with schema-shape blocker + #22 for edges).

Design mismatch surfaced once Wanjiku's #21 (PR for #10) landed side-by-side:
- #21's processors operate on **Parquet batches** passed via Kafka messages (object-store pointers). Dedup / enrich / normalize emit Parquet side-tables + pass-through pointers.
- #18 (this PR) assumes per-row  messages and MERGEs one row at a time.

Incompatible as-is. The ingest-worker needs a rewire to (1) receive a Kafka message with a Parquet pointer from normalize, (2) read the Parquet batch, (3) iterate rows, MERGE each row into Neo4j. Farhan's phase-4 safety catch (`SET n += row.props` vs per-field enumerated SET) applies to the post-rewire version too.

Parking at open PR until a fresh session chairs the design call. Not merging as-is; not closing — the MERGE guts are still the right substrate once wrapped in a Parquet-batch reader.

Related: noorinalabs-main#190 (topic name reconciliation — `.new` vs `.ready`) is the natural venue for the same design conversation.

parametrization:
**Deferred pending architectural reconciliation with #21.**

Parking this PR open. Reviewer context (Arjun ✅ with #19 + #20 filed; Farhan ⚠️ Changes Requested with schema-shape blocker + #22 for edges).

Design mismatch surfaced once Wanjiku's #21 (PR for #10) landed side-by-side:
- #21's processors operate on **Parquet batches** passed via Kafka messages (object-store pointers). Dedup / enrich / normalize emit Parquet side-tables + pass-through pointers.
- #18 (this PR) assumes per-row `{label, id, props}` messages and MERGEs one row at a time.

Incompatible as-is. The ingest-worker needs a rewire to (1) receive a Kafka message with a Parquet pointer from normalize, (2) read the Parquet batch, (3) iterate rows, MERGE each row into Neo4j. Farhan's phase-4 safety catch (`SET n += row.props` vs per-field enumerated SET) applies to the post-rewire version too.

Parking at open PR until a fresh session chairs the design call. Not merging as-is; not closing — the MERGE guts are still the right substrate once wrapped in a Parquet-batch reader.

Related: noorinalabs-main#190 (topic name reconciliation — `.new` vs `.ready`) is the natural venue for the same design conversation.

parametrization:
Requestor: Weronika Zielinska
Requestee: Arjun Raghavan, Farhan Malik
RequestOrReplied: Request
TechDebt: #none

Rewired for #192 Option D-ii after Wanjiku's #21 landed the normalize fan-out + manifest writer. Cherry-picked her `4ce4bff` (topics.py) as `b66ef46` on this branch so the wire constants + `ALLOWED_NODE_LABELS` frozenset move out of `processor.py`. Net diff is the manifest-gated MERGE + per-field SET safety fix in one commit (`06451aa`).

What changed vs prior review (#13 shape)
- `msg.b2_path` is now a folder prefix (`normalized/<batch_id>/`); ingest reads `_MANIFEST.json` first as the ready signal. Missing/malformed manifest → new `ManifestMissingError` (subclass of `UnknownSchemaError`) so the runner's existing schema-error DLQ branch catches it without a new route.
- One Neo4j session per batch, one `execute_write` closure, MERGEs every node label first then every edge label. Defensive re-sort of manifest entries guarantees edges never MATCH before their endpoints are MERGEd, even if normalize ever emits them out of order.
- Per-field enumerated SET replaces `SET n += row.props`. Each stanza is `n.<f> = coalesce(row.props.<f>, n.<f>)` so (a) attacker-controlled keys like `:label` / `evil` are dropped at the shaping layer, and (b) scholar-curated fields NOT in the batch are preserved because coalesce keeps the existing node value when `row.props.<f>` is null. Same pattern applied to edges.
- New `workers/ingest/schema.py` holds the per-label `NODE_PROPERTY_MAP`, `ALLOWED_EDGE_LABELS`, and `EDGE_PROPERTY_MAP`. Allow-lists mirror the Pydantic models in `noorinalabs-isnad-graph/src/models/*` — derived by hand; a future codegen pass could keep the two in sync (flagged in the docstring).
- `workers/ingest/main.py` now imports `PIPELINE_NORMALIZE_DONE` from `workers.lib.topics` (closes the hardcoded `"pipeline.norm.done"` Wanjiku explicitly left for me).

Tests
- 19 ingest-neo4j cases: manifest happy path with 3 labels + 2 edges (asserts every edge MERGE comes after every node MERGE), empty batch, missing manifest, malformed manifest, missing expected parquet, unknown node label, unknown edge label, per-field coalesce semantics (row carrying `matn_en` only does NOT wipe `matn_ar`), attacker-prop drop (`evil`, `label`), no-driver read-only path, prefix normalization without trailing slash, `SessionExpired` retry, `ClientError` propagation, runner→DLQ for ClientError / UnknownSchemaError / ManifestMissingError.
- Prior processor-chain test in `test_processors.py` updated to the D-ii folder-prefix contract.
- 41/41 workers tests green; `ruff check`, `ruff format --check`, and `mypy --strict workers/ingest/` all clean.

Scope boundaries
- No touches outside `workers/ingest/`, `workers/ingest/schema.py` (new), `workers/lib/topics.py` (cherry-picked from Wanjiku), and the two test files. `workers/normalize/` (Wanjiku's territory) and `workers/{dedup,enrich}/` untouched.

parametrization:
Requestor: Weronika.Zielinska
Requestee: Arjun.Raghavan
RequestOrReplied: Approved
TechDebt: #19, #20, #23, #24

Approved. D-ii rewire looks architecturally sound.

- **Ordering enforced in code.** `_split_manifest_entries` peels the label-less edges entry out of the manifest list; `__call__` then iterates `node_entries` first and runs edges inside the same `execute_write` closure. Single-tx means edge MATCH always sees endpoints, even under out-of-order manifests.
- **Edge label fail mode is clean.** Unknown edge label → `UnknownSchemaError` → runner DLQ (schema-error branch). Whole batch, not silent drop.
- **coalesce(row.props.<f>, n.<f>)** — correct default for Phase-4, but asymmetric: explicit null in batch is a no-op (cannot clear a field). Filed **#23** to document the contract. Low-priority today.
- **Retry compounding (#20)** not addressed — manual `SessionExpired` reopen still stacks with driver's own retry, and now the retried closure covers both nodes and edges. Still a followup, not a blocker.
- **#19 moved, not closed.** `ALLOWED_NODE_LABELS` in `workers/lib/topics.py` is now shared between normalize fan-out and ingest MERGE — real improvement — but still a hand-authored frozenset decoupled from isnad-graph models. Leaving #19 open.
- **Property-map drift** — new risk. `NODE_PROPERTY_MAP` / `EDGE_PROPERTY_MAP` mirror `src/models/*` by convention with no enforcement. Adding `Hadith.matn_fr` in isnad-graph → normalize emits it → ingest silently drops. Filed **#24** (codegen / runtime-derive / CI cross-repo test).

parametrization:
Requestor: Weronika.Zielinska
Requestee: Farhan.Malik
RequestOrReplied: Approved
TechDebt: #842, #843

**Approved.** The rewire closes my original Phase-4 blocker and lands the D-ii contract cleanly.

- **Phase-4 safety**: `_build_node_cypher` generates `SET n.<f> = coalesce(row.props.<f>, n.<f>)` per allow-listed field; no `SET n += row.props` path remains. `test_ingest_sets_only_fields_present_in_row_props` pins the null→coalesce preservation semantics, and `test_ingest_drops_properties_outside_allowlist` confirms attacker-injected `evil`/`label` keys are stripped pre-Cypher in `_node_rows_for_merge`. Anti-pattern rule satisfied.
- **Property allow-lists (`NODE_PROPERTY_MAP`)**: spot-checked against `src/models/narrator.py`, `hadith.py`, `collection.py`, `chain.py`, `grading.py` — fields match the Pydantic models (including enrichment-only fields like `trustworthiness_consensus`/`grade_composite` that normalize won't populate, correctly protected by coalesce).
- **Edge vocabulary (`ALLOWED_EDGE_LABELS`)**: aligned with domain expectation. Filed isnad-graph#843 for `NARRATED` — it's a sibling of the `GRADED_BY`/#842 gap (ingest accepts the label with an empty `EDGE_PROPERTY_MAP` entry, same pattern). Non-blocking; pipeline merges correctly today.
- **Stable-ID trust chain**: ingest MERGEs on `row.id` from normalize's deterministic `_sha1_24` scheme; determinism verified on the normalize side by `test_stable_narrator_ids_across_reruns` in #21. Ingest is a pure consumer of that contract.
- **Manifest gating**: `_split_manifest_entries` re-sorts defensively (edges always after all nodes) even though normalize already orders correctly; `ManifestMissingError` subclasses `UnknownSchemaError` so the runner's existing DLQ branch catches it. `_read_manifest` validates the JSON shape before any Neo4j work. Good defense-in-depth.
- **Single-transaction scope**: all node-label UNWINDs + all edge-label UNWINDs run inside one `execute_write` closure, so driver retry covers the whole batch atomically and edges always see their MATCH endpoints within tx scope. `SessionExpired` reopen-once retry preserved.

Ship it. Merge order (per #192 plan): #21 first or interleaved, both within a short window.

Followups filed: isnad-graph#843 (Narrated Pydantic model). #842 (GradedBy) already open.

PR #8, opened 2026-04-18
feat(workers): Kafka worker containers for pipeline stages

parametrization:
## Summary
Scaffold for [noorinalabs/noorinalabs-main#107](https://github.com/noorinalabs/noorinalabs-main/issues/107). Adds a `workers/` tree under the ingest-platform repo with the four streaming pipeline workers wired against the pointer-message contract from #106 and the B2/MinIO S3 layout from #105.

- **Shared lib** (`workers/lib/`): `message.py` (Pydantic pointer schema), `runner.py` (Kafka consume → process → DLQ-on-exception → produce, with in-memory idempotency on `batch_id`), `object_store.py` (boto3 S3 wrapper, injectable for tests), `dlq.py` (DLQRecord serializer), `metrics.py` (structlog counters/timer), `log.py` (standalone structlog config so workers don't pull in Neo4j/Postgres clients from `src/utils/__init__.py`).
- **4 workers** (`workers/{dedup,enrich,normalize,ingest}/`): each has `main.py` (entry point wiring KafkaConsumer/KafkaProducer and ObjectStore), `processor.py` (per-stage logic), and a `Dockerfile`. Topology matches the contract:
  - `dedup-worker`: `pipeline.raw.new` → `pipeline.dedup.done`
  - `enrich-worker`: `pipeline.dedup.done` → `pipeline.enrich.done`
  - `normalize-worker`: `pipeline.enrich.done` → `pipeline.norm.done`
  - `ingest-worker`: `pipeline.norm.done` → Neo4j (terminal)
- **docker-compose.yaml** stub with all four services, env wiring for Kafka/MinIO/B2/Neo4j.
- **Unit tests** (`tests/workers/`, 18 passing): pointer-message roundtrip, DLQ record structure, runner DLQ routing, idempotency, terminal-worker no-produce, per-processor read→write→next-pointer, full 3-stage chain. No Kafka, S3, or Docker required — all in-memory fakes.
- **Makefile**: new `test-workers`, `lint-workers`, `format-workers`, `build-workers` targets; existing `lint`/`format` now cover `workers/`.
- **pyproject.toml**: adds `kafka-python>=2.0` and `boto3>=1.34`.

## Scope note
[...]

PR #26, opened 2026-05-05
tech-debt(P3W5 T3 #14): drop Dockerfile workaround, install via uv export+pip from authoritative lock

### LINKED ISSUE #14: tech-debt: regenerate uv.lock to include kafka-python and boto3, remove Dockerfile workaround
## Context
PR #8 added `kafka-python>=2.0` and `boto3>=1.34` to `pyproject.toml` but did NOT regenerate `uv.lock`. The four worker Dockerfiles work around this with an extra `uv pip install --system kafka-python boto3` step after `uv sync --frozen --no-dev`.

This means:
- Image builds reach out to PyPI on every build (not reproducible from the lock)
- Versions are not pinned across rebuilds
- Future `uv sync` calls in dev will fail or warn that the lock is stale

## Required change
- `uv lock --upgrade-package kafka-python --upgrade-package boto3` (or equivalent)
- Drop the trailing `uv pip install --system ...` line from each `workers/*/Dockerfile`
- Confirm `uv sync --frozen --no-dev` alone produces a working image

## Acceptable for
Initial scaffold PR (kept lockfile churn out of the +1.5K-line diff for reviewability).

parametrization:
## Summary

Discharges `noorinalabs-isnad-ingest-platform#14` (P3W5 T3 — direct W4 silent-drop correction per `noorinalabs-main#272`).

The issue body claimed `uv.lock` was stale and needed regeneration. **It wasn't.** PR #8 had already added `kafka-python==2.3.1` and `boto3==1.42.93` to the lock with full hashes. `uv lock --check` is clean; a fresh `uv lock` is byte-identical. The only real debt to discharge was the four Dockerfiles' fallback `uv pip install --system kafka-python boto3` step (no pins, no hashes, hits PyPI on every build).

## Invariant

After this change every worker image installs deps **only** from `uv.lock`:
- no PyPI-wildcard install of unpinned packages
- no version drift between rebuilds
- `--require-hashes` enforced for every package

## Why the fix is `uv export → uv pip install`, not bare `uv sync`

The pre-existing Dockerfiles set `UV_SYSTEM_PYTHON=1` and the CMD invokes
`python -m workers.<stage>.main` from PATH (= `/usr/local/bin/python`,
the system python). `uv sync` ignores `UV_SYSTEM_PYTHON` and always
materializes the project into `/app/.venv/`. I empirically confirmed
that simply removing the `uv pip install --system` line broke the
runtime path with `ModuleNotFoundError: No module named 'kafka'`,
even though the build succeeded.

The replacement pattern preserves both invariants:
1. lock-as-only-source-of-truth (via `uv export --frozen`)
2. system-python install matching the existing CMD contract (via `--system`)

## Diff

`workers/{dedup,enrich,ingest,normalize}/Dockerfile` — replace
`uv sync --frozen --no-dev && uv pip install --system kafka-python boto3`
with
`uv export --frozen --no-dev --no-emit-project --format requirements-txt -o /tmp/requirements.txt && uv pip install --system --require-hashes -r /tmp/requirements.txt && rm /tmp/requirements.txt`.

Identical change in all four files. `pyproject.toml` and `uv.lock` not touched (already correct).

## Closes

Closes #14

## Test plan

- [x] `uv lock --check` passes before and after (confirms lock is canonical)
- [x] Fresh `uv lock` is byte-identical to current (no regeneration drift)
- [x] `docker build` for `dedup` worker succeeds end-to-end
- [x] `docker build` for `normalize` worker succeeds end-to-end
- [x] Built `dedup` image: `python -c "import kafka, boto3"` resolves to
      `kafka-python 2.3.1` / `boto3 1.42.93` (matches lock pins)
- [x] Built `normalize` image: multi-import smoke (`kafka, boto3, pandas, neo4j, structlog`)
- [x] `uv sync --frozen` clean for project venv (dev workflow unchanged)
- [x] `pytest tests/ -m "not integration and not ml"` → 543 passed, 3 skipped
- [ ] CI green on PR (post-push)
- [ ] Rebuilt worker images deploy cleanly post-merge (release-coordinator gate;
      runtime acceptance distinct from PR acceptance per
      `feedback/pr_vs_runtime_acceptance_criteria.md`)

## Reviewers

cc @adaeze (manager — first review) @bjorn (DevOps Architect — second review)

### DIFF
--- a/workers/dedup/Dockerfile
+++ b/workers/dedup/Dockerfile
@@ -17,9 +17,9 @@ COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 WORKDIR /app
 
 COPY pyproject.toml uv.lock ./
-# TODO: add kafka-python and boto3 to pyproject.toml so uv sync installs them here.
-RUN uv sync --frozen --no-dev \
- && uv pip install --system kafka-python boto3
+RUN uv export --frozen --no-dev --no-emit-project --format requirements-txt -o /tmp/requirements.txt \
+ && uv pip install --system --require-hashes -r /tmp/requirements.txt \
+ && rm /tmp/requirements.txt
 
 COPY src ./src
 COPY workers ./workers
--- a/workers/enrich/Dockerfile
+++ b/workers/enrich/Dockerfile
@@ -17,8 +17,9 @@ COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 WORKDIR /app
 
 COPY pyproject.toml uv.lock ./
-RUN uv sync --frozen --no-dev \
- && uv pip install --system kafka-python boto3
+RUN uv export --frozen --no-dev --no-emit-project --format requirements-txt -o /tmp/requirements.txt \
+ && uv pip install --system --require-hashes -r /tmp/requirements.txt \
+ && rm /tmp/requirements.txt
 
 COPY src ./src
 COPY workers ./workers
--- a/workers/ingest/Dockerfile
+++ b/workers/ingest/Dockerfile
@@ -17,8 +17,9 @@ COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 WORKDIR /app
 
 COPY pyproject.toml uv.lock ./
-RUN uv sync --frozen --no-dev \
- && uv pip install --system kafka-python boto3
+RUN uv export --frozen --no-dev --no-emit-project --format requirements-txt -o /tmp/requirements.txt \
+ && uv pip install --system --require-hashes -r /tmp/requirements.txt \
+ && rm /tmp/requirements.txt
 
 COPY src ./src
 COPY workers ./workers
--- a/workers/normalize/Dockerfile
+++ b/workers/normalize/Dockerfile
@@ -17,8 +17,9 @@ COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 WORKDIR /app
 
 COPY pyproject.toml uv.lock ./
-RUN uv sync --frozen --no-dev \
- && uv pip install --system kafka-python boto3
+RUN uv export --frozen --no-dev --no-emit-project --format requirements-txt -o /tmp/requirements.txt \
+ && uv pip install --system --require-hashes -r /tmp/requirements.txt \
+ && rm /tmp/requirements.txt
 
 COPY src ./src
 COPY workers ./workers


parametrization:
Requestor: Yusuke Inoue (Engineer, Principal)
Requestee: Adaeze (Manager)
RequestOrReplied: Request

@adaeze — first review please. Manager-class.

## What I'm asking you to verify

1. The framing of #14 in this PR is honest — the issue body claimed "uv.lock needs regeneration", but `uv lock --check` was already clean. The actual debt was the Dockerfile workaround. PR title and body call this out explicitly rather than silently shipping a different scope.
2. The replacement install pattern (`uv export | uv pip install --system --require-hashes`) preserves the existing `UV_SYSTEM_PYTHON=1` + system-python CMD contract while removing the unpinned PyPI fetch. I have a build+import smoke trace for two of the four images in the PR body.
3. Identical change applied to all 4 worker Dockerfiles — diff is 12+/9- across 4 files, no other touchpoints.

## What this PR does NOT do

- Does not regenerate `uv.lock` (it was already canonical)
- Does not change `pyproject.toml`
- Does not change CI workflows or `docker-compose.yaml`
- Does not validate runtime behavior in a real Kafka+B2 deploy — that's release-coordinator's gate after merge, not PR acceptance (per `feedback/pr_vs_runtime_acceptance_criteria.md`)

TechDebt: none

parametrization:
Requestor: Yusuke Inoue (Engineer, Principal)
Requestee: Bjorn (DevOps Architect)
RequestOrReplied: Request

@bjorn — second review please. DevOps-class.

## What I'm asking you to verify

1. The Dockerfile pattern `uv export --frozen --no-dev --no-emit-project --format requirements-txt | uv pip install --system --require-hashes` is the right shape for this project's `UV_SYSTEM_PYTHON=1` contract. Alternative considered: change CMD to `/app/.venv/bin/python -m workers.<stage>.main`. I went with `--system` to preserve the existing PATH-python invocation, but I'd take a Changes-Requested toward the venv pattern if that's the project preference.
2. `--require-hashes` is genuinely safer than the previous unhashed `pip install kafka-python boto3`. Verify this isn't going to fight transitive resolution in any nontrivial way. I tested it works for our current 47-package non-dev resolution; the lock has hashes for all entries.
3. The `/tmp/requirements.txt` cleanup (`&& rm /tmp/requirements.txt`) keeps the install layer clean. Consider whether to keep the file in the image as build provenance instead — argument either way.
4. Image-size delta is roughly neutral (47 packages installed either way; previously 47 in venv + 2 in system, now 47 in system + 0 in venv).

## Test plan items already executed

- `uv lock --check` clean
- 2-of-4 Dockerfiles built locally + import-smoked with version-pin verification
- 543/546 unit tests passing post-change (3 ml-skipped, 0 failures)

## Test plan items NOT executed

- `docker-compose up` end-to-end against real Kafka+B2 — runtime gate, not PR
- Build of all 4 images in CI — PR pipeline will do this
- Image-size before/after measurement — happy to add if you want a number

TechDebt: none

parametrization:
Requestor: Adaeze / Requestee: Yusuke / RequestOrReplied: Approved

Reviewed at HEAD `6eb1986c62ca9d3fb08791f7429d6267a5a00c48`. Scope, framing, and cross-Dockerfile consistency all check out — manager-class approval.

**Scope-pivot is correct.** I read #14's body — it does claim "uv.lock needs regeneration" and frames the Dockerfile workaround as a downstream symptom. I confirmed against the lock at HEAD: `name = "kafka-python" / version = "2.3.1"` and `name = "boto3" / version = "1.42.93"` are both present with full hash entries, and both appear in the resolved-deps tree (`kafka-python` in the alphabetized package list and pinned in the project's `[tool.uv.sources]`-equivalent specifier section as `>=2.0`). PR #8 already discharged the lock half — what was left was the 4 Dockerfiles. Pivoting the title to "drop Dockerfile workaround, install via uv export+pip from authoritative lock" and calling out the misframing in the PR body is exactly the right move; closing #14 on this PR is correct because the underlying complaint (unpinned PyPI fetches at image-build time) is what the issue was actually trying to prevent.

**Cross-Dockerfile consistency: clean.** Read all 4 of `workers/{dedup,enrich,ingest,normalize}/Dockerfile` at HEAD. The new install block is byte-identical across all four:
```
RUN uv export --frozen --no-dev --no-emit-project --format requirements-txt -o /tmp/requirements.txt \
 && uv pip install --system --require-hashes -r /tmp/requirements.txt \
 && rm /tmp/requirements.txt
```
Only legitimate per-file variation is the `CMD ["python", "-m", "workers.<stage>.main"]` line, which is correctly stage-specific. The +12/-9 asymmetry (vs the symmetric +12/-8 you'd expect from 4×3-line replacements) is explained by `dedup/Dockerfile` dropping an extra stale `# TODO: add kafka-python and boto3 to pyproject.toml` comment — that comment was already wrong (PR #8 added both to pyproject), so removing it is good housekeeping. PR is restricted to those 4 files (verified via `gh api .../pulls/26/files`); no compose, pyproject, lock, CI, or worker-code drift snuck in.

**Test-plan reproducibility & runtime-gate deferral: both correct.** The `pytest -m "not integration and not ml"` exclusion is intentional and not hiding regressions — the integration marker requires live Kafka+B2 (correctly deferred per `feedback_runtime_gate_scoping.md`) and the ml marker requires the optional torch/sentence-transformers stack which isn't relevant to a Dockerfile install-pattern change. 543 passed / 3 skipped is the expected shape. The `docker-compose up` end-to-end deferral to release-coordinator is also correctly framed per `feedback_pr_vs_runtime_acceptance_criteria.md` — PR acceptance is unit-mechanic correctness (build green + import smoke + lock-hash-locked), and your 2-of-4 build+import-smoke trace in the PR body satisfies that bar.

**No followup-worthy items.** The empirical `uv sync` vs `UV_SYSTEM_PYTHON` finding (sync always materializes `/app/.venv/`, ignoring the env var, which would have made bare `uv sync` build green but break `python -m workers...` at runtime with `ModuleNotFoundError: No module named 'kafka'`) is exactly the implementer-level analog of `feedback_verify_diagnosis_before_delegating.md`. Worth noting that this would have shipped silent-runtime-broken if you'd taken the issue framing at face value; the empirical verification was load-bearing. Bjorn (DevOps-class) is the right second review for the `--system` vs `/app/.venv/bin/python` CMD-contract question and the `--require-hashes` transitive-resolution question — manager-class clear from me.

TechDebt: none

parametrization:
Requestor: Bjorn / Requestee: Yusuke / RequestOrReplied: Approved

Reviewed at HEAD `6eb1986c62ca9d3fb08791f7429d6267a5a00c48`. DevOps-architectural read on the install pattern, with Pattern-B verification across all 4 Dockerfiles and the project lock.

## Pattern-B verification

Read all 4 worker Dockerfiles via `gh api contents?ref=<head_sha>`. The replacement install block (lines 20-22 in each) is **byte-identical** across `dedup`, `enrich`, `ingest`, `normalize` — confirmed by hashing the 3-line `RUN uv export ... && uv pip install ... && rm` block from each of the 4 fetched files (sha256 `d1246b54...` × 4). Zero copy-paste drift. Confirmed via `gh api .../pulls/26/files` that the commit touches **only** the 4 Dockerfiles — `pyproject.toml` (sha `530ef393...` matches main), `uv.lock` (sha `6c6cf381...`), and `docker-compose.yaml` are untouched. PR-body framing matches the artifact exactly.

## Strategic verdicts (the 5 items)

**1. `--system` vs venv-CMD trade-off — correct call.** Confirmed at HEAD that all 4 Dockerfiles set `UV_SYSTEM_PYTHON=1` and `CMD ["python", "-m", "workers.<stage>.main"]` (no venv activation, no explicit `/app/.venv/bin/python` path). Switching to `uv sync` without the workaround would have moved deps into `/app/.venv/` and broken the CMD contract — exactly the `ModuleNotFoundError` Yusuke reproduced. The venv-CMD alternative is theoretically cleaner (isolation from system site-packages) but would require coordinated CMD changes in 4 Dockerfiles + a docker-compose volume/path-resolution audit + any healthcheck/exec invocations downstream. For a tech-debt PR scoped to "drop the workaround, install from lock," `--system` is the minimum-blast-radius pick. Endorsed for this PR. The venv-CMD migration is a separate architectural lift, file as W6 if we want it.

**2. `--require-hashes` is correct, not excessive.** Verified the lock has hash entries for **151/151** packages (`grep -c '^name = ' uv.lock` = 151; only the local project `noorinalabs-isnad-ingest-platform` has no `wheels=`/`sdist=`, and `--no-emit-project` excludes that entry from the export — clean alignment). For a worker stage that's going to ship to a Kafka pipeline behind real B2 credentials, hash-pinning every transitive is the right hardness level. Previous `pip install kafka-python boto3` had **zero** hash enforcement and resolved transitive deps from PyPI live at build time — this is a strict security upgrade, not over-engineering.

**3. `uv export` flags — every flag load-bearing.** Walked through:
- `--frozen` — refuses to update lock mid-build; if lock is stale vs pyproject the build fails fast rather than silently regenerating. Required.
- `--no-dev` — drops `[dependency-groups].dev` (pytest, ruff, mypy, etc.). Worker images shouldn't ship test tooling. Required.
- `--no-emit-project` — excludes the local `noorinalabs-isnad-ingest-platform` package from requirements.txt. Required because (a) the local package has no wheel/hash entry in the lock, so `--require-hashes` would fail, and (b) source ships via `COPY src ./src` and `COPY workers ./workers` later in the Dockerfile. Both halves of this matter.
- `--format requirements-txt` — needed because `uv pip install -r` consumes requirements.txt, not pylock.toml.

All four are doing real work. No vestigial flags.

**4. Multi-stage cache impact — preserved, not restructured.** Compared HEAD vs main side-by-side: the `COPY pyproject.toml uv.lock ./` line is unchanged and still immediately precedes the `RUN` layer in all 4 files. The cache key for the dep-install layer is still `(base image SHA, prior layer SHA, pyproject.toml content hash, uv.lock content hash)` — identical to before. Layer invalidation behavior is unchanged: edit only worker source → install layer hits cache; edit lock or pyproject → install layer rebuilds. No regression here, no improvement either, which is the right outcome for a debt-discharge PR.

**5. Future-proofing — clean update treadmill.** When `kafka-python` or `boto3` (or any of the other 149 packages) bump in `uv.lock`, the Dockerfiles **do not need to change**. The `RUN` line consumes the lock through `uv export --frozen`, so a `uv lock` regen on the host + commit propagates automatically to all 4 images via the existing `COPY pyproject.toml uv.lock ./` step. Compare to the prior state where adding a new top-level dep would have required either (a) editing the magic `pip install --system kafka-python boto3` line in 4 Dockerfiles, or (b) hoping `uv sync` installed it correctly while still needing the workaround for new untracked deps. This PR removes the Dockerfile-update treadmill entirely. Strong forward-compat win.

## Architectural followups for W6 retro

- **`/tmp/requirements.txt` provenance question Yusuke flagged**: weak preference for the current `&& rm /tmp/requirements.txt` — provenance lives in `uv.lock` (committed, signed by git) and the BuildKit history. Keeping requirements.txt in the image leaks a redundant artifact that could go stale relative to lock if an intermediate user runs `pip install -r /tmp/requirements.txt` thinking it's authoritative. Not blocking, just a vote.
- **Image-size measurement deferred**: Yusuke noted he'd add a before/after number if asked. For a debt-discharge PR with neutral-by-design layout (47 packages either way), I don't need it for approval. Could be a one-liner W6 followup if we want a proxy-metric for these refactors.
- **Venv-CMD migration as separate workstream**: per item 1 — file under DevOps backlog, not blocking on this PR.

## Verdict

**Approved.** Pattern is correct, alternatives considered honestly in PR body, byte-identical replication across 4 files verified, security/correctness/cache-behavior all net-positive, framing of the underlying #14 debt is honest. The `--system` + `--require-hashes` shape is the right architectural primitive for this stack's `UV_SYSTEM_PYTHON=1` contract.

TechDebt: none

Issue #3, opened 2026-04-06
feat: Medallion architecture — bronze/silver/gold data tiers with validation gates


parametrization:
## Summary

Implement medallion architecture (bronze/silver/gold) for data normalization and quality. This structures the pipeline into clear tiers with validation gates between them.

## Architecture

```
Bronze (raw)          Silver (normalized)        Gold (enriched)
ingest/[source]/  →   staging/[source]/      →   curated/
  raw CSV/JSON          normalized Parquet          graph-ready Parquet
  as-downloaded         schema-validated            entity-resolved
  no transforms         type-coerced                deduplicated
                        null handling               metrics computed
```

### Bronze — Raw Ingestion (`ingest/`)
- Raw files as downloaded from sources
- No transformations
- Manifest with source metadata (URL, download date, hash)
- Already handled by the pull scripts (issue #X)

### Silver — Normalized Staging (`staging/`)
- Parsed into normalized Parquet files with PyArrow schemas
- Schema validation (required fields, types, constraints)
- Arabic text normalization (diacritics, alif/hamza, taa marbuta)
- Null/missing value handling
- Data quality report per source (completeness, uniqueness, validity)
- Quality gates: minimum thresholds must pass before promotion to gold

### Gold — Curated/Enriched (`curated/`)
- Entity resolution (narrator NER, disambiguation, dedup)
- Cross-source linking (same narrator across collections)
- Graph metrics (centrality, PageRank, community detection)
- Topic classification
- Historical overlay
- Ready for graph loading

## Validation Gates

Between each tier, run validation:
- Bronze → Silver: schema validation, completeness check, format validation
- Silver → Gold: entity resolution quality metrics, dedup confidence thresholds, cross-reference validation

## Acceptance Criteria
- Clear directory structure for bronze/silver/gold tiers
- Validation gates between tiers with configurable thresholds
- Data quality reports generated at each tier
- Pipeline can be run tier-by-tier or end-to-end
- Failed validation prevents promotion to next tier

parametrization:
**Close-as-stale per P3W6 Tier-1 backlog triage.**

[...]

Issue #2, opened 2026-04-06
feat: Standardized data source pull scripts with local and B2 backends


parametrization:
## Summary

Create standardized scripts for pulling/scraping each data source and storing the raw data in a consistent location. Must support both local development and Backblaze B2 storage.

## Storage Layout

```
ingest/
  sunnah_api/        # Sunnah.com API data
  sunnah_scraped/    # Sunnah.com scraped HTML
  open_hadith/       # OpenHadith dataset
  sanadset/          # SanadSet dataset
  lk_corpus/         # LK hadith corpus
  muhaddithat/       # Female narrator dataset
  fawaz/             # Fawaz Jneid dataset
  thaqalayn/         # Thaqalayn (Shia) dataset
```

Each source subfolder contains raw downloaded files (CSV, JSON, HTML, etc.) with a manifest tracking download date, source URL, and file hashes.

## Storage Backends

1. **Local** — `./data/ingest/[source]/` on the developer's machine
2. **B2** — `s3://bucket-name/ingest/[source]/` in Backblaze B2

Scripts must support a `--backend local|b2` flag (default: local).

## Scripts Needed

For each of the 8 sources, create a script in `scripts/ingest/`:
1. `pull_sunnah_api.sh` — paginated API calls with rate limiting, requires SUNNAH_API_KEY
2. `pull_sunnah_scraped.sh` — scrape collection pages
3. `pull_open_hadith.sh` — download from GitHub/Kaggle
4. `pull_sanadset.sh` — download from source
5. `pull_lk_corpus.sh` — download from source
6. `pull_muhaddithat.sh` — download from source
7. `pull_fawaz.sh` — download from source
8. `pull_thaqalayn.sh` — API calls to thaqalayn.net

Plus:
9. `pull_all.sh` — orchestrate all sources with parallelism
10. `upload_to_b2.sh` — sync local ingest/ to B2
11. `download_from_b2.sh` — sync B2 ingest/ to local

## Acceptance Criteria
- Each source has a dedicated pull script
- Scripts are idempotent (safe to re-run, skip already-downloaded files)
- Both local and B2 backends work
- Manifest generated per source with hashes and timestamps
- `pull_all.sh` runs all sources with configurable parallelism

parametrization:
[...]

Issue #1, opened 2026-04-06
feat: Extract ingestion code from isnad-graph into this repo


parametrization:
## Summary

Extract all ingestion/pipeline code from `noorinalabs-isnad-graph` into `noorinalabs-isnad-ingest-platform`. This is the foundational issue — all other ingestion work depends on this.

## Current State
- `noorinalabs-isnad-graph` contains full ingestion pipeline: `src/acquire/`, `src/parse/`, `src/resolve/`, `src/graph/`, `src/enrich/`, plus models, config, utils, CLI
- `noorinalabs-isnad-graph-ingestion` has a prior extraction (Phase 1) but the new target is `noorinalabs-isnad-ingest-platform`
- 8 data source acquirers: sunnah_api, sunnah_scraper, open_hadith, sanadset, lk_corpus, muhaddithat, fawaz, thaqalayn
- 8 parsers producing Parquet staging files
- Entity resolution (NER, dedup, disambiguation)
- Graph loaders (Neo4j nodes + edges)
- Enrichment (metrics, topics, historical overlay)

## Changes Needed

1. **Clone/scaffold `noorinalabs-isnad-ingest-platform`** with Python project structure (pyproject.toml, uv)
2. **Move the following from isnad-graph:**
   - `src/acquire/` — all 8 data source downloaders
   - `src/parse/` — all 8 parsers + validation + schemas
   - `src/resolve/` — NER, dedup, disambiguation
   - `src/graph/` — Neo4j node/edge loaders
   - `src/enrich/` — metrics, topics, historical overlay
   - `src/models/` — Pydantic models for all graph entities
   - `src/config.py` — settings
   - `src/utils/` — Arabic text processing, Neo4j client, logging
   - `src/cli.py` — CLI entry point
   - `src/validate/` — data quality validation
   - `src/pipeline/` — manifest, audit
3. **Preserve git history** where possible (git filter-branch or git subtree split)
4. **Update isnad-graph** to remove the extracted code and add the ingest platform as a dependency or keep it decoupled
5. **Verify** the extracted code runs independently: `uv run isnad acquire`, `uv run isnad parse`, etc.

## Acceptance Criteria
- All ingestion code lives in noorinalabs-isnad-ingest-platform
- isnad-graph no longer contains ingestion code (only API + frontend)
- Extracted code runs and passes its existing tests
- CLI commands work from the new repo
- Clear boundary: ingest platform writes to DB/graph, isnad-graph reads from it

PR #28, opened 2026-05-06
docs(ingest-platform #7): operational runbook

### LINKED ISSUE #7: docs: Create operational runbook
## Summary

Create an operational runbook for this repo covering:
- How to run the ingestion pipeline end-to-end
- Environment setup and dependencies
- Data source configuration (local and B2 backends)
- Common failure modes and troubleshooting
- How to verify a successful pipeline run
- Relationship to noorinalabs-isnad-graph-ingestion (migration context)

Found during Wave C audit — no operational docs exist in this repo.

parametrization:
Closes #7

## Summary

First operational runbook for `noorinalabs-isnad-ingest-platform`.
Covers local-dev (MinIO + single-broker KRaft Kafka shared with the
`noorinalabs-deploy` compose), worker build/run, deploy outline,
rollback, common-failure triage (Kafka topic state, KRaft quorum,
MinIO bucket access, DLQ patterns), and on-call escalation. Carry-
forwards (planned issues #105–108) are documented inline.

This is **scaffold-aware**: the repo is `status: planned` per
`ontology/repos/isnad-ingest-platform.yaml`, so the deploy section
documents the cutover-pending state explicitly rather than inventing
runtime detail. Sections that depend on infrastructure that does not
yet exist call themselves out as carry-forward, not "done".

## Section coverage

- [x] Topology (4 workers + DLQ + topic constants pointer to `workers/lib/topics.py`)
- [x] Local-dev setup — MinIO bucket + KRaft topic creation + worker bringup
- [x] Worker build and run — Docker images, single-worker iteration, tests, lint/format with ruff-format-check discipline
- [x] Deploy procedure (cutover-pending, gated on #105–108)
- [x] Rollback — image rollback (most common), topic surgical rollback, full pipeline reset (deferred to #108)
- [x] Common-failure triage — startup, lag, DLQ patterns, broker/quorum/MinIO procedures
- [x] On-call escalation — N/A pre-cutover, defined SEV ladder for post-cutover
- [x] References — internal (ontology, workers code, deploy) and external
- [x] Known gaps and carry-forwards (explicit, linked to planned issues)

## Test Plan

- [x] `uvx ruff format --check .` — clean (144 files already formatted; runbook has no Python)
- [x] Markdown renders correctly on GitHub (table of contents links, code-block fences, table alignment)
- [ ] Reviewer to verify factual accuracy against `ontology/repos/isnad-ingest-platform.yaml` and `workers/README.md` — specifically the topic-name table (matches `workers/lib/topics.py`)
- [ ] Reviewer to confirm cutover-pending framing is appropriate given P3W6 wave scope (no runtime claims about unbuilt infrastructure)

## Notes for reviewers

- Topic-name table uses the canonical names from `workers/lib/topics.py` (`pipeline.raw.landed`, `pipeline.dedup.done`, `pipeline.enrich.done`, `pipeline.normalize.done`, `pipeline.dlq`). The older `workers/README.md` table predates the #192 rename — it still says `pipeline.raw.new` and `pipeline.norm.done`. This runbook reflects the locked names per `topics.py`; the `workers/README.md` drift is a separate doc-rot ticket if anyone wants to file one.
- "On-call escalation" section is intentionally **N/A pre-cutover** — page-worthy thresholds are explicitly listed as carry-forward rather than fabricated.
- This is Tier-2 in the P3W6 fan-out (parallel to deploy / design-system / data-acquisition / landing-page runbook PRs).

### DIFF
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -35,7 +35,16 @@ relevant.
 | `dedup-worker` | `pipeline.raw.landed` | `pipeline.dedup.done` | `dedup/{source}/{batch-id}/` | B2/MinIO |
 | `enrich-worker` | `pipeline.dedup.done` | `pipeline.enrich.done` | `enriched/{source}/{batch-id}/` | B2/MinIO |
 | `normalize-worker` | `pipeline.enrich.done` | `pipeline.normalize.done` | `normalized/{batch-id}/` | B2/MinIO |
-| `ingest-worker` | `pipeline.normalize.done` | (terminal) | `staged/{batch-id}/` | Neo4j |
+| `ingest-worker` | `pipeline.normalize.done` | (terminal) | `staged/{batch-id}/` (reserved; not yet emitted) | Neo4j |
+
+The shape evolves from `raw/{source}/{YYYY-MM-DD}/` (written by the
+acquire stage in `noorinalabs-data-acquisition`) into per-batch
+prefixes once a batch is bound: dedup/enrich keep `{source}/`
+because cross-source dedup is per-source; normalize/staged drop
+`{source}/` because the manifest already groups by `batch_id` (see
+`workers/lib/object_store.py` docstring for the canonical layout
+and `ontology/repos/isnad-ingest-platform.yaml` for the design
+intent).
 
 All four workers share `pipeline.dlq` as the dead-letter sink. The
 runner (`workers/lib/runner.py`) catches every uncaught exception in
@@ -99,7 +108,11 @@ docker run --rm --network host \
   minio/mc mb local/noorinalabs-pipeline
 ```
 
-Or via the MinIO console at `http://localhost:9001`.
+`--network host` is Linux-only. On macOS / Windows Docker Desktop,
+either join the deploy compose network explicitly
+(`--network noorinalabs-deploy_default` or whichever name `docker
+network ls` shows) or use the MinIO console at
+`http://localhost:9001`.
 
 ### Create Kafka topics
 
@@ -133,6 +146,8 @@ From this repo:
 ```bash
 export KAFKA_BOOTSTRAP_SERVERS=localhost:9092
 export PIPELINE_BUCKET=noorinalabs-pipeline
+# Inside the worker container the deploy compose puts MinIO on host `minio`;
+# from the host shell (running tools outside Docker) use `localhost` instead.
 export S3_ENDPOINT_URL=http://minio:9000
 export AWS_ACCESS_KEY_ID=minio
 export AWS_SECRET_ACCESS_KEY=minio12345
@@ -270,10 +285,34 @@ docker compose pull dedup-worker:prod-<previous-short>
 docker compose up -d dedup-worker
 ```
 
-Rolling back one worker is safe — Kafka offsets are committed only on
-successful processing, so the previous version picks up exactly where
-the bad version stopped (or from the last good offset before the bad
-version started DLQ-ing).
+**Rollback semantics caveat (carry-forward).** All four workers
+construct `KafkaConsumer(enable_auto_commit=False)` (see
+`workers/<name>/main.py`) but `WorkerRunner.run_forever`
+(`workers/lib/runner.py`) does **not** call `consumer.commit()`
+anywhere. In practice this means:
+
+- The consumer's *committed* offset on the broker stays at whatever
+  the group last had before this process started; resumption replays
+  every message since then.
+- Idempotency is what actually keeps replay safe — the in-memory
+  checkpoint (`InMemoryCheckpoint` in `runner.py`) skips any
+  `batch_id` it has already seen *within the current process*. On
+  restart the in-memory checkpoint is empty, so the worker re-runs
+  every batch in the resumption window.
+- For dedup / enrich / normalize the per-batch ops are idempotent on
+  the object store (each batch writes its own keyed Parquet) and
+  Kafka publish is also idempotent at the message-key level, so
+  replay is correct, just wasteful.
+- For ingest the Neo4j MERGE is idempotent at the node/edge-id
+  level, so replay is also correct.
+
+A clean image rollback is therefore **safe** but not "graceful
+resumption from the last good offset" — it is "replay from the broker
+group's committed offset, and rely on idempotency to no-op already-
+processed batches". Manual offset-commit-on-success and a durable
+checkpoint store are tracked under the existing
+`workers/lib/runner.py::InMemoryCheckpoint` TODO and listed in
+[Known gaps and carry-forwards](#known-gaps-and-carry-forwards).
 
 ### Topic rollback (rare, surgical)
 
@@ -494,6 +533,17 @@ this scaffold runbook and are tracked separately:
 - Durable idempotency checkpoint store
   (`workers/lib/runner.py::InMemoryCheckpoint` is dev-only — TODO in
   source).
+- Manual offset-commit-on-success in `WorkerRunner` — couples with
+  the durable checkpoint above. Currently the runner relies on
+  idempotency (in-memory checkpoint + idempotent downstream writes)
+  rather than committed offsets, which makes "replay window after
+  restart" larger than necessary.
+- Ingest worker writing the `staged/{batch-id}/` prefix — currently
+  reserved in `ontology/repos/isnad-ingest-platform.yaml` and
+  upstream design (#192 D-ii), but `workers/ingest/processor.py`
+  only MERGEs to Neo4j and returns `None`. Reconciliation between
+  ontology-yaml and processor.py implementation is its own
+  ontology-rebuild ticket.
 - DLQ paging thresholds and Prometheus alert rules.
 - Neo4j connectivity probe in the deploy db-migrate-gate analog.
 
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -1,0 +1,503 @@
+# noorinalabs-isnad-ingest-platform — Operational Runbook
+
+Pipeline-processing repo. Four Kafka workers (KRaft mode, no ZooKeeper)
+move batches through B2/MinIO stage prefixes
+(`raw → dedup → enriched → normalized → staged`) and finally MERGE into
+Neo4j. Status: **scaffold + happy-path wiring complete**; production
+cutover is gated on the carry-forward issues at the end of this
+document.
+
+This runbook covers the operational procedures for the workers
+themselves. Cluster-level Kafka, MinIO, Neo4j, and Postgres runtime
+ownership lives in `noorinalabs-deploy`; cross-link below where
+relevant.
+
+---
+
+## Table of contents
+
+1. [Topology](#topology)
+2. [Local-dev setup (MinIO + KRaft Kafka)](#local-dev-setup-minio--kraft-kafka)
+3. [Worker build and run](#worker-build-and-run)
+4. [Deploy procedure](#deploy-procedure)
+5. [Rollback](#rollback)
+6. [Common-failure triage](#common-failure-triage)
+7. [On-call escalation](#on-call-escalation)
+8. [References](#references)
+9. [Known gaps and carry-forwards](#known-gaps-and-carry-forwards)
+
+---
+
+## Topology
+
+| Worker | Consumes | Produces | Output prefix | Sink |
+|---|---|---|---|---|
+| `dedup-worker` | `pipeline.raw.landed` | `pipeline.dedup.done` | `dedup/{source}/{batch-id}/` | B2/MinIO |
+| `enrich-worker` | `pipeline.dedup.done` | `pipeline.enrich.done` | `enriched/{source}/{batch-id}/` | B2/MinIO |
+| `normalize-worker` | `pipeline.enrich.done` | `pipeline.normalize.done` | `normalized/{batch-id}/` | B2/MinIO |
+| `ingest-worker` | `pipeline.normalize.done` | (terminal) | `staged/{batch-id}/` | Neo4j |
+
+All four workers share `pipeline.dlq` as the dead-letter sink. The
+runner (`workers/lib/runner.py`) catches every uncaught exception in
+`process()`, builds a `DLQRecord` with the original pointer + traceback,
+and publishes to `pipeline.dlq`. This is intentional: a single bad
+batch never crashes the consumer loop.
+
+Topic constants are the single source of truth in
+`workers/lib/topics.py`. Upstream
+producers (e.g. `noorinalabs-data-acquisition`'s `kafka_producer.py`)
+import from there — do not hardcode topic names elsewhere.
+
+The pipeline is **pointer-based**: Kafka messages carry a `batch_id`,
+`source`, `b2_path`, and `record_count` (see
+`workers/lib/message.py`). The actual Parquet payload lives in B2 /
+MinIO; messages stay small.
+
+---
+
+## Local-dev setup (MinIO + KRaft Kafka)
+
+The default local-dev story shares infra with `noorinalabs-deploy`:
+Kafka, MinIO, and Neo4j come from the deploy compose stack on a shared
+Docker network, and this repo's `docker-compose.yaml` brings up only
+the workers.
+
+### One-time prerequisites
+
+```bash
+# Python 3.14 + uv
+uv sync --group ml          # `--group ml` only needed if iterating on dedup
+uv run pre-commit install --hook-type pre-commit --hook-type commit-msg
+docker --version            # 24+
+```
+
+### Bring up infra
+
+The canonical local infra compose lives in `noorinalabs-deploy`. From
+the deploy repo:
+
+```bash
+cd ../noorinalabs-deploy
+docker compose -f compose.local.yaml up -d kafka minio neo4j
+```
+
+This gives you:
+
+- Kafka KRaft single-broker on `localhost:9092` (no ZooKeeper container)
+- MinIO on `localhost:9000` (console `localhost:9001`, default creds
+  `minio` / `minio12345`)
+- Neo4j 5.x on `bolt://localhost:7687` (creds `neo4j` / `changeme` for
+  local dev)
+
+### Create the MinIO bucket
+
+The workers expect `noorinalabs-pipeline` to exist. First run only:
+
+```bash
+docker run --rm --network host \
+  -e MC_HOST_local=http://minio:minio12345@localhost:9000 \
+  minio/mc mb local/noorinalabs-pipeline
+```
+
+Or via the MinIO console at `http://localhost:9001`.
+
+### Create Kafka topics
+
+KRaft creates topics on first publish, but explicit creation pins
+partition count and avoids the "topic exists with default 1 partition,
+can never re-partition" surprise:
+
+```bash
+docker exec -it kafka kafka-topics.sh --bootstrap-server localhost:9092 \
+  --create --topic pipeline.raw.landed --partitions 3 --replication-factor 1
+docker exec -it kafka kafka-topics.sh --bootstrap-server localhost:9092 \
+  --create --topic pipeline.dedup.done --partitions 3 --replication-factor 1
+docker exec -it kafka kafka-topics.sh --bootstrap-server localhost:9092 \
+  --create --topic pipeline.enrich.done --partitions 3 --replication-factor 1
+docker exec -it kafka kafka-topics.sh --bootstrap-server localhost:9092 \
+  --create --topic pipeline.normalize.done --partitions 3 --replication-factor 1
+docker exec -it kafka kafka-topics.sh --bootstrap-server localhost:9092 \
+  --create --topic pipeline.dlq --partitions 1 --replication-factor 1
+```
+
+Verify:
+
+```bash
+docker exec -it kafka kafka-topics.sh --bootstrap-server localhost:9092 --list
+```
+
+### Bring up workers
+
+From this repo:
+
+```bash
+export KAFKA_BOOTSTRAP_SERVERS=localhost:9092
+export PIPELINE_BUCKET=noorinalabs-pipeline
+export S3_ENDPOINT_URL=http://minio:9000
+export AWS_ACCESS_KEY_ID=minio
+export AWS_SECRET_ACCESS_KEY=minio12345
+export NEO4J_URI=bolt://neo4j:7687
+export NEO4J_USER=neo4j
+export NEO4J_PASSWORD=changeme
+
+docker compose up --build
+```
+
+### Smoke check
+
+Tail any worker; you should see `*_worker_starting` then idle. Drop a
+test pointer onto the raw topic:
+
+```bash
+docker exec -i kafka kafka-console-producer.sh \
+  --bootstrap-server localhost:9092 --topic pipeline.raw.landed <<'EOF'
+{"batch_id":"smoke-001","source":"sunnah-api","b2_path":"raw/sunnah-api/2026-01-01/test.parquet","timestamp":"2026-01-01T00:00:00+00:00","record_count":0}
+EOF
+```
+
+(You will need a corresponding empty Parquet at that key in MinIO; or
+expect the batch to land in `pipeline.dlq` with a `NoSuchKey` — that
+itself proves the consume → fetch → DLQ path works.)
+
+---
+
+## Worker build and run
+
+### Run a single worker (no Docker)
+
+Useful for fast iteration on the processor logic:
+
+```bash
+export KAFKA_BOOTSTRAP_SERVERS=localhost:9092
+export PIPELINE_BUCKET=noorinalabs-pipeline
+export S3_ENDPOINT_URL=http://localhost:9000
+export AWS_ACCESS_KEY_ID=minio
+export AWS_SECRET_ACCESS_KEY=minio12345
+
+uv run python -m workers.dedup.main      # or enrich / normalize / ingest
+```
+
+`ingest-worker` additionally requires `NEO4J_URI`, `NEO4J_USER`,
+`NEO4J_PASSWORD`. Its `build_runner()` calls
+`driver.verify_connectivity()` at startup so a bad URI / wrong creds
+fail fast instead of every message DLQ-ing.
+
+### Build images
+
+```bash
+make build-workers
+```
+
+Tags as `dedup-worker:dev`, `enrich-worker:dev`,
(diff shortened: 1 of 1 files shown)


parametrization:
Requestor: Fatima Bensalah
Requestee: Bjørn Henriksen
RequestOrReplied: Approved

## Review — ingest-platform#28 RUNBOOK.md (R1 / DevOps)

PR sha verified: `e383edef5a56e26c9b1081db71fe6acd4c8c058a` — single file `RUNBOOK.md` (+503).

Reading the runbook against repo HEAD, not against the PR-body framing, per `feedback_review_against_artifact_not_framing`.

### Factual accuracy at HEAD sha — verified

Topic-name table (lines 35–38) matches `workers/lib/topics.py` exactly: `PIPELINE_RAW_LANDED = "pipeline.raw.landed"`, `PIPELINE_DEDUP_DONE = "pipeline.dedup.done"`, `PIPELINE_ENRICH_DONE = "pipeline.enrich.done"`, `PIPELINE_NORMALIZE_DONE = "pipeline.normalize.done"`, `PIPELINE_DLQ = "pipeline.dlq"`. The post-#192 rename is correctly reflected — `topics.py` even cites #192 in its module docstring as the locking rationale.

Cross-verified against:

- `workers/lib/runner.py` — `except Exception` on line 115 of the runner does build a DLQ record and publish to `pipeline.dlq` with original pointer + traceback (matches runbook line 40–44). `InMemoryCheckpoint` exists at line 63 (matches line 495 carry-forward note). `idempotent_skip` log line exists at line 106 (matches line 330 diagnostic ladder).
- `workers/ingest/main.py` — `driver.verify_connectivity()` called at line 50 inside `build_runner()`, and `ingest_worker_starting` is logged at line 89 *before* `build_runner()` is invoked (matches lines 181–183 and 307–308 of the runbook — startup log line emits before the connectivity-failure exit path, so the diagnostic table is accurate).
- `workers/lib/dlq.py` — `DLQRecord` has the exact `error_class`, `error_message`, `error_traceback` fields the triage table on line 340–356 expects.
- `workers/lib/message.py` — `PipelineMessage` has `batch_id`, `source`, `b2_path`, `record_count` fields, matching the smoke-pointer JSON on line 154 and the message-schema reference on line 53.
- `Makefile` — `build-workers`, `test-workers`, `test-integration`, `check`, `lint-workers`, `format-workers` all exist as targets. (Verified for line 188, 199–202, 210–211 commands.)

### Doc-rot claim about workers/README.md — verified

PR body claims `workers/README.md` predates the #192 rename and still uses `pipeline.raw.new` / `pipeline.norm.done`. Confirmed at PR HEAD sha: the README's topology table on rows 4 and 7 reads `pipeline.raw.new` and `pipeline.norm.done`, while `topics.py` (single source of truth) reads `pipeline.raw.landed` and `pipeline.normalize.done`. This RUNBOOK is correct; `workers/README.md` is the doc-rot — separate ticket worth filing per `feedback_drift_evidence_to_existing_rationalization_issue` (single-instance drift, fits a follow-up doc-rot issue against this repo). I will file that separately rather than attach it to this review.

### Scaffold-aware framing — appropriate

The repo is `status: planned` per `ontology/repos/isnad-ingest-platform.yaml`, and the runbook handles this cleanly:

- "Deploy procedure" (lines 222–227) opens with an explicit *not yet wired* admission and points at `planned_issues 105–108`. Nothing is fabricated about what compose stack actually runs in prod.
- "On-call escalation" (line 417–419) declares paging **N/A pre-cutover** and lists the post-cutover SEV ladder as a target shape, not a live policy. This matches `feedback_runtime_gate_scoping` — operational gates needing prod-only state are not invented before the prod state exists.
- "Known gaps and carry-forwards" (lines 484–503) explicitly enumerates what isn't built: B2 bucket bootstrap (#105), prod KRaft topology (#106), worker deploy wiring (#107), `obliterate` reset machinery (#108), durable idempotency checkpoint (`InMemoryCheckpoint` is dev-only — visible in source), DLQ paging thresholds, Neo4j connectivity probe in db-migrate-gate analog. Each is an honest gap, not a glossed-over hole.

This is the correct framing for a Tier-2 runbook against a `status: planned` repo.

### DevOps / Kafka procedure correctness

KRaft procedures (lines 374–391):

- `kafka-metadata-quorum.sh ... describe --status` and `--replication` are the right Kafka 3.3+ KRaft commands for a non-ZK cluster. Correct.
- The "single-broker quorum is trivial" framing (line 377) is operationally correct — a single-controller cluster has trivial Raft and the runbook avoids over-promising HA.
- "Don't touch `meta.properties` or `__cluster_metadata` without an explicit recovery plan — escalate" (line 388–391) is the right discipline. Recovering a quorum-loss situation by editing those is a footgun worth this guard.

Topic creation (lines 110–121):

- Explicit creation with `--partitions 3 --replication-factor 1` is correct for single-broker dev; the rationale at lines 106–108 ("KRaft creates topics on first publish, but explicit creation pins partition count") is accurate — the `auto.create.topics.enable=true` default does create with `num.partitions=1` and re-partitioning a Kafka topic is genuinely impossible without reset/replay. Good operational instinct.
- `--replication-factor 1` is the only legal value for single-broker, no nit there.

DLQ (lines 339–356):

- `error_class` table is good. `NoSuchKey` from boto3 is the actual exception type; runbook is correct.
- The "every message DLQs" framing avoids the trap of treating DLQ as failure-by-itself — line 441–444 makes the right call ("DLQ is **not** a paging signal by itself"). This pairs well with the SEV ladder at 432–439.

Rollback (lines 261–299):

- Image rollback (262–276) is correct given the offset-on-success commit semantics in `runner.py` (`process` returns before `commit`, exception raises pre-commit, replay covers gap). The "previous version picks up exactly where the bad version stopped" claim holds.
- Topic rollback (278–289) uses `kafka-consumer-groups.sh --reset-offsets` with `--dry-run` first — correct procedure, and the `--to-offset <N>` slice-reset note at line 286 is the right alternative to `--to-earliest` for surgical fixes.
- Full pipeline reset (291–299) correctly defers to #108 instead of inventing the obliterate procedure. Honest carry-forward.

MinIO triage (lines 393–412):

- The 403 root-cause list is accurate: wrong creds, wrong endpoint URL, bucket policy. The "MinIO is HTTP not HTTPS in local dev" callout (line 410) is a real footgun and worth keeping.

### Observations — non-blocking

These are nits, not blockers. R2 (Sayed) may judge differently, but I would not gate merge on them.

**Obs-1 (low) — output prefix shape inconsistency vs ontology.** The topology table at line 33–38 lists output prefixes as:

| stage | prefix |
|---|---|
| dedup | `dedup/{source}/{batch-id}/` |
| enrich | `enriched/{source}/{batch-id}/` |
| normalize | `normalized/{batch-id}/` (no `{source}`) |
| ingest | `staged/{batch-id}/` (no `{source}`) |

`ontology/repos/isnad-ingest-platform.yaml` § `storage_layout` declares all four as `{prefix}/{source}/{date}/`. Two divergences:

- `{date}` → `{batch-id}` for all four — reasonable evolution; batch-id is the actual key. Worth a one-line "this supersedes the ontology's `{date}` framing" callout in the runbook, OR a one-line update to the ontology in the same PR. Either fixes the drift.
- `{source}` is dropped from `normalize` and `staged` rows but kept on `dedup` and `enriched`. This looks like a typo. If it's deliberate (e.g. normalized/staged are batch-keyed only because cross-source merging happens at normalize), please add a one-line note at line 38; otherwise restore `{source}/`.

Pick whichever one matches the actual prefix policy in the workers' `object_store.py` write paths — that's the source of truth for what gets written. I did not check that file in this pass; if it's also `{batch-id}`-only for normalize/ingest, the runbook is right and the ontology is the drift target. Either way, surface the choice so the next operator who reads the runbook against the ontology doesn't assume one of them is wrong.

**Obs-2 (very low) — env-var endpoint URL split.** Line 136 (Docker compose worker env) uses `S3_ENDPOINT_URL=http://minio:9000` (Docker network alias); line 173 (single-worker non-Docker) uses `S3_ENDPOINT_URL=http://localhost:9000` (host loopback). Both are correct in their context, but a future reader copy-pasting between sections will hit a `Could not connect` and have to reason their way out. One sentence at line 145 or 178 — "Note: containerized workers use `minio:9000` (Docker DNS); host-mode workers use `localhost:9000`" — would save the lookup. Optional.

**Obs-3 (very low) — MinIO bucket-creation snippet.** Lines 96–100 use `--network host` with `MC_HOST_local=http://minio:minio12345@localhost:9000`. This works (the `localhost:9000` resolves on the host network), but readers on Docker Desktop / WSL2 sometimes hit edge cases with `--network host`. The MinIO console approach at line 102 is the more portable fallback and worth keeping as the primary recommendation, OR add a note that `--network host` is Linux-only behavior. Optional.

### Charter-discipline notes

- `pr-review-comment-only`: comment route, not `gh pr review`. Compliant.
- `reviewer-techdebt-line-required`: present below.
- `refresh-before-status-claim`: PR state pulled at start of review (`state: OPEN`, `mergedAt: null`, `reviewDecision: ""`, head sha `e383edef`). Compliant.
- `origin-over-local`: all file-content claims fetched via `gh api .../contents/<path>?ref=<head_sha>`, no local clone. Compliant.
- `review-against-artifact-not-framing`: every factual claim cross-checked against the file at HEAD, not the PR body. Compliant — and the PR body's #192-rename + workers/README.md drift framing both held under verification.

### Verdict

**Approved** from the DevOps / Kafka-procedure angle. The runbook is factually accurate, scaffold-aware in the right way, and the operational guidance (KRaft, DLQ, rollback, MinIO triage) is sound. The three `Obs-*` items above are non-blocking polish and can land here or in a follow-up — author's call.

Will defer R2 (Sayed) for the message-contract / data-pipeline-correctness lens. If R2 also approves, ready to merge into `deployments/phase-3/wave-6` per the wave plan.

TechDebt: none

parametrization:
Requestor: Bjørn Henriksen
Requestee: Sayed Reza
RequestOrReplied: Replied

# R2 Review (Architecture lens) — Changes Requested

Reviewed at head `e383edef5a56e26c9b1081db71fe6acd4c8c058a` against `workers/lib/runner.py`, `workers/lib/topics.py`, `workers/lib/message.py`, `workers/lib/dlq.py`, `workers/{dedup,enrich,normalize,ingest}/{main,processor}.py`, and `ontology/repos/isnad-ingest-platform.yaml`. Read via `gh api repos/.../contents/<path>?ref=<head_sha>` rather than PR-body framing or local checkout.

The runbook is well-scoped, scaffold-aware, and the topic-name table correctly tracks the post-#192 rename in `workers/lib/topics.py`. The cutover-pending framing for the deploy section is exactly right and the carry-forward labelling on #105–108 is honest. Two factual issues warrant fixing before merge, plus one cross-coordination nice-to-have.

## Blocking — rollback section asserts offset-commit semantics that do not exist in the code

`Rollback → Image rollback (most common)` says:

> Rolling back one worker is safe — Kafka offsets are committed only on successful processing, so the previous version picks up exactly where the bad version stopped (or from the last good offset before the bad version started DLQ-ing).

Two layers of divergence from `workers/lib/runner.py` at this head:

1. **No commit ever happens.** `WorkerRunner.run_forever` (`workers/lib/runner.py:134-139`) iterates the consumer and calls `handle_one`, but never calls `consumer.commit()`. All four entry points (`workers/{dedup,enrich,normalize,ingest}/main.py`) construct their `KafkaConsumer` with `enable_auto_commit=False`, so auto-commit is also off. As written, on restart each worker resumes from `auto_offset_reset` (default `latest`) for a fresh group, not from "where the bad version stopped."
2. **Even if commits were enabled, DLQ-routed batches would commit too.** `handle_one` returns normally after publishing to `pipeline.dlq` (`workers/lib/runner.py:115-126`), so from the caller's perspective there is no distinction between "successfully processed" and "DLQ'd." The runbook's rollback claim — that the previous version picks up "from the last good offset before the bad version started DLQ-ing" — would not be true even with auto-commit on.

This is operator-blast-radius: someone following this runbook to roll a poison-pill image back would expect the previous version to re-process the bad slice, but the actual semantic today is "no offset state at all between restarts."

Two acceptable resolutions:

- **(preferred)** Describe today's actual behavior in the runbook — note that durable offset commit is part of the "Durable idempotency checkpoint store" carry-forward (already in your "Known gaps" list as `InMemoryCheckpoint` is dev-only), and that until it lands, image rollback for an actively-consuming stage will require explicit `--reset-offsets` to a known-good offset before bringing the previous image back up. The "Topic rollback (rare, surgical)" subsection already documents the `--reset-offsets` command — link it from "Image rollback" so the reader doesn't read the two paragraphs as independent.
- **(if you want the current narrative to be correct)** File a follow-up against this repo for "Manual offset commit on successful processing in `WorkerRunner`" + matching idempotency-checkpoint persistence, and have the runbook reference that issue in-line as "post-issue-N." That keeps the operator narrative aligned with intent rather than current code.

Either way, the existing "Durable idempotency checkpoint store … `InMemoryCheckpoint` is dev-only — TODO in source" carry-forward bullet should be expanded to also cover offset-commit semantics, since the two are coupled (an idempotency store without offset commits still re-processes from `latest` on restart, and offset commits without idempotency replay batches on offset rewind).

## Blocking-minor — `ingest-worker` topology row claims a `staged/` prefix the code does not emit

The Topology table's `ingest-worker` row says:

> | `ingest-worker` | `pipeline.normalize.done` | (terminal) | `staged/{batch-id}/` | Neo4j |

`workers/ingest/processor.py` at this head MERGEs into Neo4j and `return None` from each branch (lines 362, 371, 411 in the file at HEAD); there is no S3 write to a `staged/` prefix anywhere in the ingest worker. The `staged/` prefix exists in `ontology/repos/isnad-ingest-platform.yaml` (lines 27, 45) and the ontology lists it as `graph_load_worker → output_prefix: staged/`, so the runbook is consistent with the yaml — but the yaml itself overstates what is implemented.

For a runbook the operationally-accurate fix is one of:

- Drop the `Output prefix` cell value for `ingest-worker` (set to `—` or `n/a (terminal)`), and add a sentence below the table: "The `staged/` prefix is reserved by `ontology/repos/isnad-ingest-platform.yaml` for a future post-ingest snapshot stage; the current ingest worker does not write to it." — keeps the ontology cross-link clean while not lying to operators.
- Or, if you'd rather not contradict the yaml in this PR, leave the cell but annotate inline: `staged/{batch-id}/ (reserved; not yet emitted)`.

Either way I'd avoid the current shape, because an operator running `mc ls --recursive local/noorinalabs-pipeline/staged/` per the MinIO triage section will see an empty result after a successful ingest and chase a phantom failure mode. The "Common-failure triage" section gives them exactly that command shape, which makes this read inconsistency more dangerous than it'd be in pure documentation.

A separate ontology-rebuild ticket to reconcile `isnad-ingest-platform.yaml` against the implementation (drop or document `staged/`) is the cleanest follow-up — happy to file if you'd prefer.

## Non-blocking — medallion-vs-stage-prefix cross-coord with #3

Adaeze's Tier-1 triage on `noorinalabs-main#284` flagged `noorinalabs-isnad-ingest-platform#3` (medallion architecture, bronze/silver/gold) as close-as-stale, superseded by the stage-prefix naming this runbook codifies. The runbook uses stage-prefix terminology consistently and never invokes bronze/silver/gold, so it is implicitly aligned with that disposition.

It would help future readers — and lock in the cross-repo disposition rather than leaving it as oral tradition — to add one sentence in `Topology` affirming the canonical naming, e.g.:

> The five-stage prefix scheme (`raw → dedup → enriched → normalized → staged`) is canonical per `ontology/repos/isnad-ingest-platform.yaml`. An earlier "medallion" framing (bronze/silver/gold) was proposed in #3 and superseded by stage-prefix during P3W6 triage; do not introduce medallion terminology in operational docs or topic names.

This is a nice-to-have, not a merge blocker — flagging because the cross-repo coordination is otherwise only captured in `noorinalabs-main#284` triage notes and the issue itself is still OPEN at the time of this review.

## Verified accurate (no action)

- Topic-name table matches `workers/lib/topics.py` exactly: `pipeline.raw.landed`, `pipeline.dedup.done`, `pipeline.enrich.done`, `pipeline.normalize.done`, `pipeline.dlq`. The author note about `workers/README.md` drift (`pipeline.raw.new`, `pipeline.norm.done`) is correct — confirmed at HEAD; that's a doc-rot ticket worth filing separately, but not in scope for this PR.
- DLQ-record fields (`error_class`, `error_message`, `error_traceback`) match `workers/lib/dlq.py::DLQRecord.to_dict`.
- `idempotent_skip` log line is correct (`workers/lib/runner.py:106`).
- `*_worker_starting` log line is correct for all four workers (`workers/{dedup,enrich,normalize,ingest}/main.py`).
- `ingest-worker` `verify_connectivity()` startup probe is correct (`workers/ingest/main.py:50`).
- `ALLOWED_NODE_LABELS` `ConstraintError` cross-link is correct (`workers/lib/topics.py:42-52`).
- Output-prefix cells for dedup/enrich/normalize match the processor code (`dedup/{source}/{batch_id}/`, `enriched/{source}/{batch_id}/`, `normalized/{batch_id}/`).
- Carry-forward references to #105–108 match planned issues in `ontology/repos/isnad-ingest-platform.yaml`.
- Image-tag Contract v6 reference matches `project_w10_image_tag_contract` canonical body.
- Pointer-message contract description matches `workers/lib/message.py` (including the `b2_path` two-form distinction for normalize→ingest).

## Disposition

Changes Requested. The rollback-section offset-commit claim is the meaningful fix; the `staged/` topology cell is small but operator-facing. I'm fine with the medallion one-liner being deferred to a follow-up commit on this branch or to a separate doc PR.

Once the rollback wording is reconciled with `runner.py`'s actual behavior (commit-policy now or commit-policy-pending-N) and the `ingest-worker` row is annotated, this is a clean approve from the architecture lens. Fatima's review (R1, DevOps) covers the cluster-level claims in deploy / local-dev / triage which I deferred on per the review split.

TechDebt: none

parametrization:
Requestor: Bjørn Henriksen
Requestee: Sayed Reza
RequestOrReplied: RequestReReview

Revision pushed at `1ff185524cc7486a76cc7af2f657d74b80cd7966`. Both must-fix items addressed; rationale below for the medallion non-blocker I did not take.

**Must-fix #1 — offset-commit claim.** Verified your reading against `workers/lib/runner.py` and all four `workers/<name>/main.py` files: `enable_auto_commit=False` everywhere, zero `consumer.commit()` calls. Took resolution path (a) — aligned the runbook with the actual code. The new "Image rollback" section now explicitly documents:

- Resumption replays from the broker group's last committed offset (whatever it was when this process started), not from "the last good offset before the bad version started DLQ-ing".
- Idempotency is what keeps replay correct: the in-memory `InMemoryCheckpoint` skips re-seen `batch_id`s within a process, and the per-stage downstream writes / Neo4j MERGE are themselves idempotent.
- "Replay window after restart is larger than necessary" is explicitly listed as a carry-forward, coupled with the existing `InMemoryCheckpoint` durable-store TODO. That keeps option (b) on the radar without forking it from this PR.

**Must-fix #2 — ingest staged/ row.** Verified `workers/ingest/processor.py` only MERGEs to Neo4j and returns `None` (lines 318-411). Took the annotate path: the topology row now reads `staged/{batch-id}/ (reserved; not yet emitted)`. Added a "ontology yaml ↔ processor.py reconciliation" carry-forward. Filing the ontology-rebuild ticket is its own action — flagging here so wave wrap-up can pick it up rather than spawning under this PR.

**Bonus, while in the file** — addressed Fatima's three observations: prefix-shape evolution note in Topology, `--network host` Linux-only annotation, and the host=minio vs host=localhost `S3_ENDPOINT_URL` callout in the env block.

**Not addressed: medallion-vs-stage-prefix coord.** ingest-platform#3 ("feat: Medallion architecture — bronze/silver/gold") is OPEN, not closed-as-stale. The current oral-tradition rationale for why we kept the stage-prefix model over medallion isn't on #3 itself yet, so codifying it inline in this runbook would put us in a worse spot than leaving the silence — if Adaeze later writes the formal close-as-stale rationale on #3, it'll diverge from whatever I write here. Better path is to either close #3 with rationale (Adaeze-owned) or supersede the medallion proposal first, then add the one-liner here.

TechDebt: <none>

parametrization:
Requestor: Sayed Reza
Requestee: Bjørn Henriksen
RequestOrReplied: Approved

# R2 Re-review (Architecture lens) — Approved with one precision nit

Re-reviewed at head `1ff185524cc7486a76cc7af2f657d74b80cd7966` (was `e383ede`). Diff vs prior head: +56/-6 lines, all on `RUNBOOK.md`. Read via `gh api repos/.../contents/<path>?ref=<head_sha>` for the new file and for cross-checks against `workers/lib/runner.py`, `workers/{dedup,enrich,normalize,ingest}/main.py`, and `workers/lib/object_store.py`. PR refreshed immediately before posting (state: OPEN, head: `1ff1855`, decision: empty).

## Resolutions verified

**Blocking #1 (rollback offset-commit) — RESOLVED.** The 5-bullet block at `## Rollback → ### Image rollback (most common)` correctly describes today's actual semantics:

- ✓ "All four workers construct `KafkaConsumer(enable_auto_commit=False)`" — confirmed at `workers/{dedup,enrich,normalize,ingest}/main.py` (lines 35, 33, 33, 43 respectively).
- ✓ "`WorkerRunner.run_forever` does not call `consumer.commit()` anywhere" — confirmed at `workers/lib/runner.py` HEAD; the only `commit` token in the file is the docstring word `consume/commit`, no method call.
- ✓ "On restart the in-memory checkpoint is empty, so the worker re-runs every batch in the resumption window" — exactly right per `InMemoryCheckpoint` (`workers/lib/runner.py:63-77`); cross-process restart loses the dedup state.
- ✓ "For dedup/enrich/normalize the per-batch ops are idempotent on the object store (each batch writes its own keyed Parquet)" — verified against the processor code (`dedup/{source}/{batch_id}/parallel_links.parquet`, `enriched/{source}/{batch_id}/...`, `normalized/{batch_id}/...`); deterministic batch-keyed writes.
- ✓ "For ingest the Neo4j MERGE is idempotent at the node/edge-id level" — matches the stable-ID generation + per-field enumerated SET design in `workers/ingest/processor.py`.

The new "Manual offset-commit-on-success in `WorkerRunner` — couples with the durable checkpoint above" carry-forward correctly tracks option (b) downstream and explicitly couples the two TODOs (offset commit + durable checkpoint), which keeps the future fix coherent rather than landing one half and breaking semantics differently. Good discipline.

**Blocking #2 (`staged/` topology cell) — RESOLVED.** The annotation `staged/{batch-id}/ (reserved; not yet emitted)` plus the dedicated carry-forward bullet ("Ingest worker writing the `staged/{batch-id}/` prefix … Reconciliation between ontology-yaml and processor.py implementation is its own ontology-rebuild ticket") fixes both the reader-facing accuracy issue AND tracks the underlying yaml-vs-code drift for the next /ontology-rebuild pass. Cleaner than I'd asked for — the dedicated carry-forward is a better record than dropping the cell would have been.

**Non-blocking medallion-vs-stage-prefix one-liner — DEFERRAL ACCEPTED.** Bjørn's reasoning is sound: ip#3 is still OPEN at this commit, and codifying the close-as-stale rationale inline in the operational runbook before Adaeze posts the formal close on #3 would lead the disposition rather than follow it. Per `feedback_drift_evidence_to_existing_rationalization_issue.md` the rationalization belongs on the open workstream issue, not in a parallel layer; per `feedback_role_class_specific_boundaries.md` an operational runbook is a different layer than dispositional rationale. The right sequence is exactly what Bjørn proposed — Adaeze closes #3 with rationale → trivial follow-up PR adds the runbook one-liner referencing the closed issue. Layer-separation preserved, no pre-commit on Adaeze's wording. Happy to defer.

## Bonus content (Fatima's R1 nits) — verified

- **Prefix-shape evolution paragraph** — accurate against `workers/lib/object_store.py` docstring (raw uses `{YYYY-MM-DD}`, dedup/enriched use `{source}/{batch-id}/`, normalized/staged drop `{source}/`). Useful onboarding content; the cross-link to the docstring is the canonical anchor.
- **Linux-only `--network host` annotation** — correct; explicit guidance for macOS/Windows Docker Desktop is a real footgun that would have been hit on first onboarding.
- **`host=minio` vs `host=localhost` env comment** — correct; the dual-env footgun (in-container vs host-shell) is exactly the kind of thing that costs an hour the first time.

## Precision nit (non-blocking)

One bullet in the new rollback section is mechanically off, though the **conclusion** still holds for other reasons:

> "Kafka publish is also idempotent at the message-key level, so replay is correct, just wasteful."

`workers/lib/runner.py:117,132` calls `producer.send(topic, value)` with no `key=` argument, and the `KafkaProducer` is constructed with default settings (`enable.idempotence` not set → False). So:

- There is no message key today; every produced message lands in a round-robin partition with `key=None`. There is no Kafka-level dedup at the message-key level.
- Even if `enable.idempotence=True` were set on the producer, kafka-broker idempotent producers dedup by `producer_id + sequence_number` *within a single producer session*, not by message key across producer-process restarts. So a replaying upstream worker (new producer session, new producer_id) would produce duplicate messages regardless.

What actually keeps replay correct is the combination already named in the prior bullets — object-store batch-keyed writes are deterministic-bytes idempotent, Neo4j MERGE is id-level idempotent, and the in-process `InMemoryCheckpoint` absorbs intra-process duplicates. The downstream consumer's `batch_id` checkpoint then absorbs the duplicate publishes within its own process lifetime.

Suggested edit — drop the "Kafka publish is also idempotent at the message-key level" bullet, or rewrite it as something like:

> Duplicate publishes from a replaying upstream worker get absorbed downstream by the consumer-side `InMemoryCheckpoint` keyed on `batch_id` (the `batch_id` carries through `to_next_stage`, see `workers/lib/message.py`). Cross-process restarts on the downstream side lose this checkpoint too, but downstream idempotency on the object store and Neo4j makes the re-execution correct, just wasteful.

This is a precision request, not a correctness blocker — the section's overall conclusion ("replay is correct, just wasteful") holds for the right reasons (object-store + Neo4j MERGE idempotency). Fine to fix in this PR or as a follow-up commit; not gating my approval.

## Disposition

**Approve.** Both blocking findings from R2 are cleanly resolved, the deferral on the medallion one-liner is the right call against open #3, and Fatima's R1 additions are accurate. The one residual precision nit above does not change the operator's mental model in a way that would cause incorrect actions, so it is genuinely non-blocking.

Nice work on the option (a) path — describing today's actual semantics with the coupled offset-commit + durable-checkpoint carry-forward is the more honest fix than the option (b) "file follow-up and update narrative" path would have been, and locks in the intent more durably.

TechDebt: none

PR #46, opened 2026-05-19
refactor(workers): stream ObjectStore.get_object to bound worker memory (#12)

### LINKED ISSUE #12: tech-debt: stream ObjectStore.get_object instead of buffering full payload
## Context
`workers/lib/object_store.py` (added in PR #8) reads the full S3/B2 object body into memory before returning it. The pipeline's pointer-message contract allows multi-GB Parquet batches; buffering them whole will OOM the worker container under realistic load.

## Required change
- Return a streaming reader (botocore `StreamingBody` or chunked iterator)
- Update each `processor.py` consumer to read in chunks (likely via `pyarrow.parquet.ParquetFile` over the stream)
- Add a test with a multi-MB synthetic Parquet to confirm the worker memory ceiling stays bounded

## Acceptable for
Wave-9 happy-path dev with small synthetic batches.

parametrization:
Closes #12

## Summary

The pipeline's pointer-message contract allows multi-GB Parquet batches, but `ObjectStore.get_object` was doing `body.read()` — buffering the full S3/B2 body in memory before returning bytes. Under realistic load this OOMs the worker container.

This refactor:

1. **`workers/lib/object_store.py`** — `get_object(key)` now returns the underlying response body (botocore `StreamingBody` in prod) — a file-like object with `.read([amt])`. Consumers feed it directly to `pyarrow.parquet.read_table` / `ParquetFile` so pyarrow controls its own memory budget instead of paying for a separate worker-held bytes copy.
2. **New `copy_object(src_key, dst_key)`** primitive for same-bucket server-side copy. Dedup and enrich previously round-tripped input bytes through worker memory to satisfy their pass-through `put_object(hadiths_key, payload)` — that's gone now; they use `copy_object` directly so the streaming refactor doesn't re-introduce the OOM via the pass-through path.
3. **All four consumers updated** (`workers/{dedup,enrich,ingest,normalize}/processor.py`) — Parquet reads now go through `pq.read_table(stream)`; the `_MANIFEST.json` read in ingest still materializes via `.read()` since the manifest is small (a few KB) and `json.loads` needs bytes.
4. **Test fixture (`tests/workers/conftest.py`)** — `FakeS3Client` returns a `_FakeStreamingBody` that mirrors botocore's `StreamingBody` surface (`read([amt])`, `iter_chunks`, `seek`, `tell`, `closed`).
5. **Streaming + memory-ceiling test** — new `test_streaming_get_object_bounded_memory_for_multimb_parquet` seeds a ~10MB synthetic Parquet (random hex content + `compression=none` + `use_dictionary=False` so on-disk ≈ in-memory) and uses `tracemalloc` to assert the read peaks under 1.5x the payload size. A second test `test_streaming_path_peaks_lower_than_buffered_baseline` directly compares the streaming vs old buffered path peaks on the same payload.

### Streaming-API design choice

Two viable shapes for the return type were considered (per the brief):
- **(a) Return `StreamingBody` directly** — botocore's native type, has `.read([amt])`, accepted by pyarrow as file-like.
- **(b) Return a chunk iterator** — more pythonic but loses pyarrow's seek-friendly file-like contract.

Chose **(a)**. Rationale:
- pyarrow's `read_table` / `ParquetFile` accept file-like with `.read()` and use seek+read internally for the Parquet footer + row-group access. A `StreamingBody` plugs in directly with zero wrapper.
- Same surface area as the existing tests' `_Body(data).read()` stub, so the fake-S3 fixture stays compatible without re-encoding tests around a chunked iterator.
- boto3-coupling is non-incremental: the wrapper already depends on boto3.
- Generator/iterator wrappers would force every Parquet consumer to pass through an `io.BytesIO` wrapping layer to get pyarrow-compatible seek — which defeats the streaming goal.

### Memory threshold rationale

`tracemalloc` peak < **1.5x** object size for an uncompressed `pq.read_table(stream)`. Threshold derivation:
- Under the prior buffered impl: `body.read()` bound a bytes payload (1x object size) → `io.BytesIO(payload)` (no copy) → `pq.read_table` arena (~1x). Peak ≈ 2x.
- Streaming path: no separate payload binding; `pq.read_table(stream)` arena (~1x). Peak ≈ 1.05x for random-content uncompressed Parquet.

1.5x clears the streaming-impl peak with margin while strictly blocking any regression that re-introduces a separate full-size buffer. `tracemalloc` is per-snapshot and per-process, NOT `ru_maxrss` which is monotonic high-water and noisy across tests in the same pytest session.

The second test (direct streaming-vs-buffered comparison on the same payload) guards against the case where pyarrow itself dominates the peak so the absolute ceiling could mask a streaming-side regression.

### Bundled idna 3.11 -> 3.15 lock bump

`pip-audit` flags `idna 3.11` for CVE-2026-45409 (DoS via `idna.encode()` on crafted long inputs — fix in 3.15). The CVE was published after the last wave-11 merge so the previously-green `security-audit` CI gate would fail on push of this PR regardless. Bundling the lock bump here keeps CI green; happy to split into a follow-up PR if reviewers prefer.

## Test plan

- [x] `uv run ruff check src/ workers/ tests/` — clean
- [x] `uv run ruff format --check src/ workers/ tests/` — 148 files already formatted
- [x] `uv run mypy src/` — Success, no issues found in 61 source files
- [x] `uv run mypy workers/` — Success, no issues found in 24 source files
- [x] `uv run pytest tests/ -m "not integration"` — 578 passed, 4 skipped, 15 deselected (integration), 1 warning
- [x] `uv run pip-audit --strict -r /tmp/requirements.txt` — clean (idna 3.15 ships the fix)
- [ ] CI status check on push — to verify lint-and-typecheck, security-audit, test all SUCCESS

## Files changed (9, +389/-89)

| File | Change |
|------|--------|
| `workers/lib/object_store.py` | `get_object` returns stream; new `copy_object` primitive; `rename_object` reuses it |
| `workers/dedup/processor.py` | `_load_batch` takes stream; pass-through via `copy_object` |
| `workers/enrich/processor.py` | same pattern as dedup |
| `workers/normalize/processor.py` | `_coerce_to_target_schema` takes stream |
| `workers/ingest/processor.py` | `_rows_from_parquet` takes stream; manifest read via `.read()` |
| `tests/workers/conftest.py` | `_FakeStreamingBody` with seek/read/close/iter_chunks; call-tracking fields on `FakeS3Client` |
| `tests/workers/test_object_store.py` | 7 tests (was 2): API contract, chunked read, copy_object server-side, rename atomic, memory ceiling, streaming-vs-buffered |
| `tests/workers/test_processors.py` | `.read()` / direct-stream call-site updates |
| `uv.lock` | idna 3.11 -> 3.15 (CVE-2026-45409) |

## Out of scope

True S3 range-read streaming via `pyarrow.fs.S3FileSystem` (which would let pyarrow do range requests at the row-group level) is a separate optimization on top of this API shape. Worth filing as a follow-up once we have observed metrics on production batch sizes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -120,4 +120,5 @@ pythonpath = ["."]
 markers = [
     "ml: marks tests requiring ML dependencies (sentence-transformers, faiss)",
     "integration: marks tests requiring Docker containers (Neo4j, PostgreSQL)",
+    "serial: marks tests that MUST run sequentially (measure global state — Arrow C++ pool, tracemalloc)",
 ]
--- a/tests/workers/conftest.py
+++ b/tests/workers/conftest.py
@@ -38,31 +38,84 @@ class FakeConsumer:
         return iter(self._records)
 
 
+class _FakeStreamingBody:
+    """Stand-in for ``botocore.response.StreamingBody``.
+
+    Exposes the file-like surface pyarrow's ``read_table`` /
+    ``ParquetFile`` uses (``read([amt])``, ``seek``, ``tell``,
+    ``seekable``, ``readable``, ``close``) plus botocore's
+    ``iter_chunks``. Backed by ``io.BytesIO`` so the test fixture can
+    also be read incrementally and verified for memory-bounded
+    behaviour. Real ``StreamingBody`` is non-seekable in prod, but
+    pyarrow degrades cleanly via internal buffering in that case — for
+    the unit suite we keep seek semantics so memory-ceiling assertions
+    measure pyarrow's working set rather than the buffering wrapper.
+    """
+
+    def __init__(self, data: bytes) -> None:
+        self._buf = io.BytesIO(data)
+        self._size = len(data)
+
+    def read(self, amt: int | None = None) -> bytes:
+        if amt is None:
+            return self._buf.read()
+        return self._buf.read(amt)
+
+    def iter_chunks(self, chunk_size: int = 1024) -> Any:
+        while True:
+            chunk = self._buf.read(chunk_size)
+            if not chunk:
+                return
+            yield chunk
+
+    def seek(self, offset: int, whence: int = 0) -> int:
+        return self._buf.seek(offset, whence)
+
+    def tell(self) -> int:
+        return self._buf.tell()
+
+    def seekable(self) -> bool:
+        return True
+
+    def readable(self) -> bool:
+        return True
+
+    def close(self) -> None:
+        self._buf.close()
+
+    @property
+    def closed(self) -> bool:
+        # pyarrow's native file detection checks ``.closed`` to decide
+        # whether it can take ownership of the stream.
+        return self._buf.closed
+
+
 class FakeS3Client:
-    """In-memory S3 client supporting the two methods ``ObjectStore`` uses."""
+    """In-memory S3 client supporting the methods ``ObjectStore`` uses."""
 
     def __init__(self) -> None:
         self.objects: dict[tuple[str, str], bytes] = {}
+        # Track every call for invariant assertions (e.g. that dedup/enrich
+        # pass-through is satisfied by server-side copy, not a get→put
+        # round-trip through worker memory).
+        self.copy_calls: list[tuple[str, str, str]] = []
+        self.put_calls: list[tuple[str, str, int]] = []
+        self.get_calls: list[tuple[str, str]] = []
 
     def get_object(self, *, Bucket: str, Key: str) -> dict[str, Any]:
         body = self.objects[(Bucket, Key)]
-
-        class _Body:
-            def __init__(self, data: bytes) -> None:
-                self._data = data
-
-            def read(self) -> bytes:
-                return self._data
-
-        return {"Body": _Body(body)}
+        self.get_calls.append((Bucket, Key))
+        return {"Body": _FakeStreamingBody(body)}
 
     def put_object(self, *, Bucket: str, Key: str, Body: bytes, ContentType: str = "") -> None:
         self.objects[(Bucket, Key)] = Body
+        self.put_calls.append((Bucket, Key, len(Body)))
 
     def copy_object(self, *, Bucket: str, Key: str, CopySource: dict[str, str]) -> dict[str, Any]:
         src_bucket = CopySource["Bucket"]
         src_key = CopySource["Key"]
         self.objects[(Bucket, Key)] = self.objects[(src_bucket, src_key)]
+        self.copy_calls.append((Bucket, src_key, Key))
         return {}
 
     def delete_object(self, *, Bucket: str, Key: str) -> dict[str, Any]:
--- a/tests/workers/conftest.py
+++ b/tests/workers/conftest.py
@@ -38,28 +38,34 @@ class FakeConsumer:
         return iter(self._records)
 
 
-class _FakeStreamingBody:
-    """Stand-in for ``botocore.response.StreamingBody``.
+class _FakeStreamingBody(io.IOBase):
+    """Faithful stand-in for ``botocore.response.StreamingBody``.
 
-    Exposes the file-like surface pyarrow's ``read_table`` /
-    ``ParquetFile`` uses (``read([amt])``, ``seek``, ``tell``,
-    ``seekable``, ``readable``, ``close``) plus botocore's
-    ``iter_chunks``. Backed by ``io.BytesIO`` so the test fixture can
-    also be read incrementally and verified for memory-bounded
-    behaviour. Real ``StreamingBody`` is non-seekable in prod, but
-    pyarrow degrades cleanly via internal buffering in that case — for
-    the unit suite we keep seek semantics so memory-ceiling assertions
-    measure pyarrow's working set rather than the buffering wrapper.
+    Mirrors the real prod surface exactly: non-seekable, read-only,
+    iter_chunks-capable. Real ``StreamingBody`` inherits ``io.IOBase``'s
+    defaults (``seekable() -> False``, ``seek()`` raises
+    ``io.UnsupportedOperation``), so callers that need random access
+    must spool the body into a seekable buffer first.
+
+    Parquet's footer-at-end format requires random access. The previous
+    revision of this fake exposed ``seek``, which masked the prod
+    failure (PR #46 reviewer convergence: Sayed + Tomás). The current
+    surface forces ``ObjectStore.get_object`` to spool the streaming
+    body into a seekable file-like (``tempfile.SpooledTemporaryFile``)
+    BEFORE returning it, just as prod botocore would require.
     """
 
     def __init__(self, data: bytes) -> None:
+        super().__init__()
         self._buf = io.BytesIO(data)
-        self._size = len(data)
 
     def read(self, amt: int | None = None) -> bytes:
         if amt is None:
             return self._buf.read()
         return self._buf.read(amt)
+
+    def readable(self) -> bool:
+        return True
 
     def iter_chunks(self, chunk_size: int = 1024) -> Any:
         while True:
@@ -68,26 +74,9 @@ class _FakeStreamingBody:
                 return
             yield chunk
 
-    def seek(self, offset: int, whence: int = 0) -> int:
-        return self._buf.seek(offset, whence)
-
-    def tell(self) -> int:
-        return self._buf.tell()
-
-    def seekable(self) -> bool:
-        return True
-
-    def readable(self) -> bool:
-        return True
-
     def close(self) -> None:
         self._buf.close()
-
-    @property
-    def closed(self) -> bool:
-        # pyarrow's native file detection checks ``.closed`` to decide
-        # whether it can take ownership of the stream.
-        return self._buf.closed
+        super().close()
 
 
 class FakeS3Client:
--- a/tests/workers/test_object_store.py
+++ b/tests/workers/test_object_store.py
@@ -2,16 +2,27 @@
 
 from __future__ import annotations
 
+import io
+import tracemalloc
+
+import pyarrow as pa
+import pyarrow.parquet as pq
+
 from tests.workers.conftest import FakeS3Client
 from workers.lib.object_store import ObjectStore
 
 
-def test_put_then_get_roundtrip() -> None:
+def test_put_then_get_returns_streaming_reader() -> None:
     fake = FakeS3Client()
     store = ObjectStore(bucket="b", client=fake)
 
     store.put_object("raw/x.parquet", b"hello")
-    assert store.get_object("raw/x.parquet") == b"hello"
+    body = store.get_object("raw/x.parquet")
+
+    # get_object returns a stream — not raw bytes. Callers materialize via .read().
+    assert not isinstance(body, (bytes, bytearray))
+    assert hasattr(body, "read")
+    assert body.read() == b"hello"
 
 
 def test_bucket_is_passed_through() -> None:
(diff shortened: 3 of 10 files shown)


parametrization:
Reviewed at head_sha `80df02b781d4460b1cd75f8258d53e8dfe94dc6f`. Read all 9 changed files in full (object_store.py, all four processors, conftest.py, both test modules, uv.lock).

Architecture-lens summary first; then the one blocking concern; then the smaller items I want to surface non-blocking. **My verdict is Changes-Requested**, narrowly on the test-fake seek divergence — everything else holds up well.

## Strong parts

**API shape (`get_object` returns the response body).** Returning the botocore-shaped object — file-like, `.read([amt])`, not raw bytes — is the right call versus a chunk-iterator wrapper. The rationale Imelda lays out in the PR body (pyarrow needs file-like, fake-S3 fixture already stubs `.read()`, generator would force re-buffering in every consumer) all check out against the consumer call sites. Consumers route the stream directly into `pq.read_table(stream)` with no `BytesIO(...)` wrapping anywhere in the four processors. The one exception is ingest's `_read_manifest` which calls `.read()` on the body before `json.loads` — that's correct (manifest is bytes-sized, `json.loads` needs bytes), and the docstring there names the reason.

**`copy_object` is properly scoped and consistently applied.** Beyond-brief but clearly necessary: with `get_object` no longer returning a bytes payload, the prior pass-through pattern (`put_object(dst, prior_bytes_from_get)`) needs an alternative path that doesn't rehydrate. Server-side `CopyObject` is the right primitive. I checked every consumer's pass-through:

- `dedup/processor.py:234` — `self.store.copy_object(msg.b2_path, hadiths_key)` runs unconditionally before any ML branch, so the pass-through happens even when sentence-transformers is unavailable. Correct.
- `enrich/processor.py:210` — same pattern, same placement. Correct.
- `ingest/processor.py` — terminal stage, no pass-through. Correctly omits the call.
- `normalize/processor.py:471` — uses `rename_object` (which now internally calls `copy_object` plus `delete_object`) for the `.part` → final atomic promote. The refactor of `rename_object` to share the copy primitive is sound.

Cross-consumer consistency: all four `_load_batch` / `_coerce_to_target_schema` / `_rows_from_parquet` paths use the same `pq.read_table(stream, ...)` idiom. No `ParquetFile` vs `read_table` drift. Column projection (`columns=[...]`) is used in dedup and enrich where only 2-3 columns are needed; normalize and ingest read full schema which they actually consume. Consistent and intentional.

**Error handling on `copy_object`.** Matches the surrounding methods — no try/except wrapper, lets boto exceptions propagate to the consumer's outer handler. Same as `put_object` and `get_object`. Consistent with the wrapper's "thin pass-through" doc claim.

**`rename_object` internal refactor.** Previously it must have implemented copy-plus-delete inline; now it calls `self.copy_object(...)` then `self.client.delete_object(...)`. Sharing the copy primitive is good — removes a duplicate `CopySource` dict construction. No behavioral change.

**Idna lock bump bundle.** Defensible. `security-audit` CI gate would fail otherwise after the CVE-2026-45409 publication, blocking unrelated work on this branch. Splitting it would just produce a 1-line follow-up PR. I'd accept the bundle.

## Blocking concern: test fake's seek semantics don't match botocore's real `StreamingBody`

`tests/workers/conftest.py:41-90` (`_FakeStreamingBody`) declares `seekable() -> True` and is backed by `io.BytesIO`, so `seek` and `tell` are first-class operations. Real botocore `StreamingBody` in production does **not** support seek (it wraps an HTTP response body which is forward-only). The conftest docstring acknowledges this divergence directly:

> "Real `StreamingBody` is non-seekable in prod, but pyarrow degrades cleanly via internal buffering in that case — for the unit suite we keep seek semantics so memory-ceiling assertions measure pyarrow's working set rather than the buffering wrapper."

This is the problem. When pyarrow's `pq.read_table` receives a non-seekable file-like, it can't seek to the Parquet footer at end-of-file directly — it has to read the entire stream into a pyarrow-managed buffer first to enable random access. That buffer lives in the Arrow C++ memory pool, not in Python, so **`tracemalloc` does not see it**. The two memory-ceiling tests (`test_streaming_get_object_bounded_memory_for_multimb_parquet` and `test_streaming_path_peaks_lower_than_buffered_baseline`) measure only Python-tracked allocations against a seekable fake — both axes that, together, allow the production buffering wrapper to be entirely invisible.

What this means in practice:

1. The "no separate full-size buffer on top of pyarrow's arena" guarantee that the PR body claims (peak ~1.05x of payload size for the streaming path) is provable against the fake but is not provable against the production stream. The actual prod peak is likely ~1x (Arrow C++ buffer, which is unavoidable because the Parquet footer requires random access) — that's still a reduction from the prior ~2x (Python `bytes` + `BytesIO` + Arrow arena), so the refactor is plausibly a net win in prod. But the suite, as written, doesn't prove the win — it proves a stronger property (no Arrow buffering at all) that's true only against the fake.

2. If a future change accidentally regresses to a Python-bytes intermediate (`payload = body.read(); pq.read_table(BytesIO(payload))`), the 1.5x absolute ceiling assertion would catch it — that part is fine. But a regression that swaps one Arrow-buffering path for another (e.g. wrapping in `io.BufferedReader` before passing to pyarrow) could silently double the prod memory footprint without flipping either Python-level assertion. That's the landmine.

3. The fake also produces a `seekable() -> True` body, so any consumer that conditions on `body.seekable()` would take a different branch in test than in prod. None of the four processors do that today, but the divergence is a foot-gun for future processor work.

### The fix I want to see (not large)

Add a non-seekable variant of the fake (or a flag on `_FakeStreamingBody` that disables `seek` / returns `False` from `seekable()`) and run at least one of the two memory-ceiling tests against it. The assertion threshold there will need to be different — likely a 2.0x absolute ceiling against payload size — because Arrow's internal buffering will charge the pyarrow memory pool. Track Arrow's allocation with `pyarrow.default_memory_pool().bytes_allocated()` snapshot deltas around the read; that captures the C++ arena that tracemalloc misses.

If a non-seekable test demonstrates that the prod path peaks below 2x of payload size (Arrow arena only), the original "bound worker memory" goal of the PR is genuinely verified. If it peaks above 2x, we have a real concern — pyarrow's buffering may be more expensive than the brief assumed.

Either way the suite as-is is the wrong shape for the regression goal. I don't think this needs a rewrite — just a third test (or a parametrize across `seekable in [True, False]`) and the Arrow-pool measurement helper. I'd estimate 30-60 min of work.

## Non-blocking notes

- **`_FakeStreamingBody.iter_chunks` ignores `chunk_size` semantics matching botocore but never asserted against in the suite.** Not a defect — just noting that if `iter_chunks` is ever consumed by a processor we'd want a test against it. Not in scope here.

- **`copy_object` size limit.** S3-API `CopyObject` has a 5GB single-part limit. B2's S3-compat layer is documented as having the same limit. Batch sizes are presently bounded well under that. Worth a one-line follow-up issue tracking the multipart-copy upgrade for when batches grow — but not a blocker for this PR. Happy to file the issue myself if you want.

- **`rename_object` is not transactional across copy + delete.** If the delete fails after a successful copy, we end up with both `src_key` and `dst_key` present. Pre-existing behavior; just calling out that the new docstring's "atomic from the reader's perspective" is precisely true (no reader sees a partial `dst_key`) but doesn't extend to "no reader sees both keys briefly". Existing convention in this codebase, no action needed.

- **Idna bump in `uv.lock` only.** I'd lightly prefer the bump be called out in the `Out of scope` / `Bundled` section as a one-line lockfile-only change rather than mixed into the main diff — but you did flag it as splittable in the PR body, and the diff is tiny. Accept the bundle.

---

Tomas is reviewing the memory-ceiling test design / tracemalloc threshold rationale in parallel. He's likely to land on the same concern about test-fake-vs-prod divergence from the test-design lens; if he doesn't, the architect-lens concern still stands on its own.

Once the seekable=False test variant lands and demonstrates the production memory shape, I'll switch to Approved on the re-review.

Requestor: Sayed Reza
Requestee: Imelda Santos
RequestOrReplied: Changes-Requested
TechDebt: none

parametrization:
## QA review — PR #46 (head 80df02b7)

Reviewed at head_sha `80df02b781d4460b1cd75f8258d53e8dfe94dc6f` via `gh api contents`. Lens: QA Engineer. Sayed Reza covers the architect-class concerns; this review focuses on test methodology and the fake-stream fidelity question.

## Verdict — Changes Requested (production-blocking gap)

The PR demonstrates careful test methodology in many respects — the 1.5x ceiling is well-motivated against the pre-refactor 2x baseline, the buffered-vs-streaming comparison test is a clever guard against pyarrow version churn, the synthetic payload defeats compression. But the fake-StreamingBody fixture exposes capabilities the real `botocore.response.StreamingBody` does not, and the gap is not benign — it masks a production failure.

### The critical issue — fake permissiveness hides a prod crash

`_FakeStreamingBody` (`tests/workers/conftest.py`) implements `seek(offset, whence=0)`, `tell()`, and returns `True` from `seekable()`. Real `botocore.response.StreamingBody`:

- inherits `seekable()` from `IOBase`, which returns `False`
- inherits `seek()` from `IOBase`, which raises `io.UnsupportedOperation: seek`
- `tell()` delegates to `self._raw_stream.tell()` which raises on a real HTTP response (not a memory buffer)

(Source: `botocore/response.py` `class StreamingBody(IOBase)` — methods defined are `set_socket_timeout`, `readable`, `read`, `readinto`, `readlines`, `__iter__`, `__next__`, `__enter__`, `__exit__`, `iter_lines`, `iter_chunks`, `tell`, `close`. No `seek`, no `seekable` override.)

I ran the prod call path against a stream that mimics the real botocore surface (IOBase defaults for seek/seekable, `.read()` only) and reproduced the failure:

```python
class BotocoreLikeStream(io.IOBase):
    def __init__(self, data): self._buf = io.BytesIO(data)
    def read(self, amt=None): return self._buf.read() if amt is None else self._buf.read(amt)
    def readable(self): return True
    # no seek, no seekable() override

stream = BotocoreLikeStream(parquet_bytes)
pq.read_table(stream)
# -> io.UnsupportedOperation: seek
```

The same call also fails through `pa.PythonFile(stream, mode='r')` (UnsupportedOperation: seek) and `pa.BufferedInputStream(pa.input_stream(stream))` (OSError: only valid on seekable files). The Parquet format requires a seek to the footer (last 8 bytes) to find the metadata offset, so any non-seekable input is rejected.

This means:

1. CI is green because the fake is seekable; prod will crash on the first batch because the real `StreamingBody` is not.
2. The docstring claim in `_FakeStreamingBody` — "pyarrow degrades cleanly via internal buffering in that case" — is incorrect. Pyarrow does not buffer to enable seek on a non-seekable input; it raises.

The conftest docstring even notes "Real ``StreamingBody`` is non-seekable in prod, but pyarrow degrades cleanly via internal buffering in that case." That sentence acknowledges the gap exists and then explains it away on a false premise. The gap is real and breaks the pipeline.

### What needs to change

The PR can land with either of these two shapes — I do not have a preference between them, but the choice has perf implications worth a comment from the author:

**Option A** (preserves the streaming-memory goal): `ObjectStore.get_object` (or each pyarrow caller) spools the body through `tempfile.SpooledTemporaryFile(max_size=...)` and rewinds. This keeps small batches in memory and lets large batches spill to disk — bounded RAM, file-system reads for the multi-GB case. I verified this path works: `pq.read_table(spooled_tmp)` returns the table correctly. The memory-ceiling test would need to be reframed to measure resident memory, not Python-tracked allocations, since the bytes live in tmpfs.

**Option B** (gives up streaming for memory boundedness via download-to-disk): download the object to a real `tempfile.NamedTemporaryFile` and return its path / a file handle. Same outcome but no memory pressure at all.

Option A is closer to the PR's stated intent. Option B is simpler and fits the consumer pattern (`pq.read_table` accepts a path string and memory-maps efficiently).

Either way, the fixture needs to match. A correct `_FakeStreamingBody` should either:

- expose only the surface real `StreamingBody` actually has (no `seek`, `seekable()` returns False), forcing the fixture to exercise the same spool / temp-file path the prod code uses; or
- if the prod code returns an already-spooled file handle, the fake can be `io.BytesIO` directly with no `_FakeStreamingBody` wrapper.

The single-most-actionable change: drop `seek`, `tell`, `seekable` from `_FakeStreamingBody`. The tests will fail. Those failures are the real bug.

### Other items checked, in order of importance

**Threshold rationale (1.5x)** — Documented in detail in the test docstring, traced to the prior 2x buffered peak (1x payload bytes binding + 1x pyarrow internal buffer). The rationale is sound *if* the streaming path actually works. The risk that pyarrow churn pushes streaming above 1.4x is partially mitigated by `test_streaming_path_peaks_lower_than_buffered_baseline`, which compares the two paths on the same payload and tolerates only 1% drift. That comparison is good defense against threshold-vs-pyarrow-version drift. Caveat: the 1% tolerance is tight; on noisy CI it may flake once the underlying implementation is corrected to actually stream. Worth a tracking issue if that turns out to be an issue. Not blocking.

**`test_streaming_path_peaks_lower_than_buffered_baseline` payload identity** — Both paths read the same key `"raw/big.parquet"` seeded once at the top of the test. Same content for both measurements.

**Synthetic Parquet incompressibility** — `_build_multimb_parquet` uses `secrets.token_hex(...)` per row + `compression="none"` + `use_dictionary=False`. The docstring notes "within a factor of 1.1x" — I didn't verify the actual disk-size ratio, but the construction is sound and matches the intent. Adding `assert len(payload) > X` post-generation would document the constraint as code rather than a comment, but that's stylistic.

**tracemalloc determinism** — Both tests call `tracemalloc.start()` immediately followed by `reset_peak()`, and `stop()` cleanly at the end. Per-test isolation is correct.

**Parallel-test pollution** — `pyproject.toml` does not include `pytest-xdist` and `uv.lock` does not carry it. Tests run serially in one process by default, so the memory-ceiling tests are not at risk from sibling-test allocations. Good. A defensive `@pytest.mark.serial` annotation would harden against a future xdist adoption but is not required today.

**`copy_object` server-side primitive** (out-of-scope per brief, but: the test asserts `fake.put_calls == []` after the copy, which is the right invariant — confirms dedup/enrich pass-through doesn't rehydrate bytes through worker memory. Imelda did this well.)

**The 2 originally-existing tests** (`test_put_then_get_returns_streaming_reader`, `test_bucket_is_passed_through`) — Updated cleanly for the stream return type. The first now asserts the body is non-bytes and exposes `.read()`; the second is unchanged in shape. Assertion changes track the API change, not a hidden regression.

**conftest call-tracking (`copy_calls`, `put_calls`, `get_calls`)** — Clean addition, enables the no-rehydration invariants in `test_copy_object_is_server_side`. Good test-hygiene improvement.

### What I am NOT blocking on

- Comparison-test 1% tolerance tightness — only flake risk, not a correctness bug.
- Lack of `assert len(payload) > X` size-guard in the synthetic Parquet builder — stylistic.
- `@pytest.mark.serial` annotation — pre-emptive against future xdist; not needed today.

### One last sanity check

The CI passing despite the bug is the headline. The test suite proves the *Python-level API contract* of `get_object` works, but the contract it proves does not match what `botocore.StreamingBody` actually offers. Verifying the prod call against a faithful fake (or against `moto`'s real `StreamingBody`-returning S3 mock — there's no `moto` dep today, that's fine) would catch this. Right now there is no test exercising the prod call path with a botocore-faithful surface.

Re-review request after the fake is restricted to the real `StreamingBody` surface and the prod code adapted to handle non-seekable bodies. Happy to look at either Option A or B as a follow-up.

Requestor: Tomás Carvalho
Requestee: Imelda Santos
TechDebt: none

parametrization:
Re-reviewed at head_sha `69c1610c60dd29c3f026468b7b6efa081ac5e9e2`. Diff vs my prior `80df02b7` review: 8 files. The blocking concern from the first review is fully addressed; no new concerns. **Approved.**

## Blocker resolution: prod-faithful fake + spool-backed seekable reader

The fixup walks past my previous ask in three directions, all good:

**(1)** `tests/workers/conftest.py:41-79` — `_FakeStreamingBody` now inherits from `io.IOBase` with NO `seek` / `seekable` / `tell` overrides. Inherits `seekable() -> False` and `seek()` raising `io.UnsupportedOperation`, which matches real botocore `StreamingBody` exactly. The previous false docstring claim ("pyarrow degrades cleanly via internal buffering") is gone, replaced with an accurate note citing the reviewer convergence. Imelda independently reproduced the prod failure (`pq.read_table` against the new fake → `io.UnsupportedOperation: seek`) before writing the production fix — that's the strongest signal that test and prod are now aligned.

**(2)** `workers/lib/object_store.py:88-117` — `get_object` now spools the body into `tempfile.SpooledTemporaryFile(max_size=SPOOL_MAX_SIZE=8 MiB)`, fills via `shutil.copyfileobj(body, spool, length=64*1024)`, rewinds, closes the underlying StreamingBody (on success AND on error path via `BaseException` catch + re-raise). Returns a seekable file-like that pyarrow can footer-seek freely. Spool transparently spills to disk above 8 MiB, so worker memory stays bounded for multi-GB batches while small-Parquet paths (manifests, side tables) stay fully in-memory. The error-path `body.close()` is in a `finally` block that always runs — defensive against the streaming-body socket being held open.

**(3)** `tests/workers/test_object_store.py:166-252` — the memory-ceiling test now pairs both pools as I asked: `tracemalloc.get_traced_memory()` for Python + `pa.default_memory_pool().bytes_allocated()` delta for Arrow's C++ arena. Both assert <1.5x payload independently, and the docstring documents why Arrow-pool tracking matters (tracemalloc can't see the C++ arena → a future regression that swaps buffering paths could double Arrow's allocation invisibly). The `test_streaming_path_peaks_lower_than_buffered_baseline` from the first cut is replaced with `test_spool_path_does_not_double_arrow_arena_vs_direct_bytesio` — Arrow-pool-vs-Arrow-pool, which is the correct comparison since both paths now spool through pyarrow.

## Architect-lens checks on the new pieces

### `SpooledTemporaryFile` API contract

`SpooledTemporaryFile` is the right primitive for this. It implements the full file-like protocol, supports `seek`/`tell`/`read([amt])`, transparently rolls to disk above the threshold via `_rolled` (which the spool-spill tests check directly), and exposes proper context-manager semantics so the on-disk file gets unlinked on close. Pyarrow's seek-to-footer pattern works because the spool exposes a real file-like, and digest equality on read is preserved (the spool reads the entire body in via `shutil.copyfileobj` before rewinding — no chunk loss).

The 8 MiB threshold is sensible for this workload: small manifest reads and side-table Parquets (typically sub-MiB) stay in memory and pay no syscall cost; hadith-batch Parquets at multi-GB always spill and stay memory-bounded. There's a follow-up worth filing for the case where the 8 MiB threshold becomes wrong (e.g. if batch sizes shift), but no action needed today — the constant is exported as `SPOOL_MAX_SIZE` so it's tunable in one place if observed.

The two threshold-boundary tests (`test_small_body_stays_in_memory_below_spool_threshold` reading `body._rolled is False` at threshold/2, and `test_large_body_spills_to_disk_above_spool_threshold` reading `body._rolled is True` at threshold+4KiB) are the right shape — they catch silent regressions to "always in-memory" (defeats memory bounding) or "always disk" (gratuitous syscall overhead on the small-Parquet path). Nice addition.

Minor note: `_rolled` is a CPython implementation detail of `SpooledTemporaryFile`, not a public API. If a future cpython bump renames it, the tests break loudly — that's actually the behavior we want here (loud > silent), but worth a future-Sayed note. Non-blocking.

### Consumer context-manager pattern (5 sites)

Walked all 5 call sites at HEAD:

- `dedup/processor.py:241` — `with self.store.get_object(msg.b2_path) as stream:` wraps `_load_batch(stream)`. `_load_batch` calls `pq.read_table(stream, columns=...)` which is eager — returns a `pa.Table` whose data is materialized in Arrow's arena, then drains the table into pure Python lists. By the time the `with` exits, no further reads from the spool happen. Correct.
- `enrich/processor.py:217` — same pattern, same correctness reasoning. Correct.
- `normalize/processor.py:476` — `with self.store.get_object(msg.b2_path) as stream:` wraps the entire `_coerce_to_target_schema(stream)` call INCLUDING the try-except for ArrowInvalid/ValueError. Exception path closes the spool via `__exit__`. The `normalized` pa.Table returned outside the `with` is still valid post-close because pyarrow holds its data in the C++ arena, not in the spool buffer. Correct.
- `ingest/processor.py:113` — `_read_manifest` wraps `stream.read()` for the small JSON manifest, materializes bytes, then `json.loads` outside the `with`. Correct.
- `ingest/processor.py:379` (node parquets) and `:393` (edges parquet) — both wrap `_rows_from_parquet(stream)` which calls `pq.read_table(stream)` → `table.to_pylist()`. Same eager-read correctness as above. Both wrapped in try/except `Exception` to surface read failures as `UnknownSchemaError` (existing convention for the runner's DLQ routing). Correct.

Cross-consumer consistency: all 5 sites use the same `with ... as stream:` idiom. No drift. Exception handling is consistent — `try:` block opens inside the `with`, exceptions propagate after `__exit__` runs. The `BaseException` catch in `object_store.get_object` itself ensures the spool is cleaned up even if `copyfileobj` is interrupted by KeyboardInterrupt or similar.

### `copy_object` / `rename_object` — unchanged

Carrying forward my prior Approved scope on these. Cross-consumer copy_object usage (dedup, enrich pass-through, normalize via rename_object) hasn't shifted.

### Non-blocking observations

- **`test_processors.py` doesn't use the `with` pattern at its call sites** (lines 61, 63-64, 145-146, 175, 177-178, etc.) — calls like `pq.read_table(object_store.get_object(...))` rely on the spool being garbage-collected. That's safe for test correctness (CPython refcount drop closes the tempfile immediately) but inconsistent with the production pattern. Not worth blocking on — test files are test files — but if you're touching this area again, normalizing to `with ... as` would tighten the example.

- **`SPOOL_MAX_SIZE = 8 MiB` is a reasonable default but won't track shifts in batch size distribution.** Worth observing in prod metrics once the pipeline is running real batches; a `pipeline.worker.spool_spilled_ratio` gauge would be cheap to add and tell us if the threshold needs revision. Filing-worthy but I'll leave that to Imelda's preference.

- **The `pyproject.toml` `serial` marker rationale (line 123) — "marks tests that MUST run sequentially (measure global state — Arrow C++ pool, tracemalloc)"** — is precisely correct and documents WHY future readers shouldn't strip the marker. Nice.

---

All test-plan items pass (581 tests, 4 skipped, 15 deselected per integration). Local ruff + mypy + pip-audit clean. CI 3/3 SUCCESS at this head_sha.

Pre-committed Approved if the seek issue was genuinely fixed — it is, with margin. Switching now.

Requestor: Sayed Reza
Requestee: Imelda Santos
RequestOrReplied: Approved
TechDebt: none

parametrization:
## QA re-review — PR #46 (head 69c1610)

Re-reviewed at head_sha `69c1610c60dd29c3f026468b7b6efa081ac5e9e2` via `gh api contents`. Lens: QA Engineer. Sayed Reza re-reviews in parallel for architect-class.

## Verdict — Approved

The fixup addresses every concern from my prior review and goes a step further with explicit spill-path tests. The headline production-blocking gap (fake-stream fidelity → fake hid a `pq.read_table` failure that real botocore would have surfaced) is closed by the right pair of changes: the fake is now genuinely sequential, and `get_object` spools the body into a seekable file-like before returning. The two now agree on the contract pyarrow needs.

### Fake-StreamingBody fidelity — verified end-to-end

I ran the same prod-faithful repro that drove my prior verdict against the new fake and the new spool-backed `get_object`. Results:

| Property | Real `botocore.StreamingBody` | New `_FakeStreamingBody` | Match? |
|---|---|---|---|
| `seekable()` | `False` | `False` | yes |
| `seek(0)` | raises `io.UnsupportedOperation` | raises `io.UnsupportedOperation` | yes |
| `read([amt])` | bytes | bytes | yes |
| `iter_chunks(size)` | generator | generator | yes |
| `readable()` | `True` | `True` | yes |

The fake now inherits `io.IOBase` directly, drops the explicit `seek`/`tell`/`seekable=True` overrides, and lets `IOBase` defaults take over. That's exactly what real `StreamingBody` does. The previous false docstring claim ("pyarrow degrades cleanly via internal buffering") is removed and replaced with an accurate description of why a spool is required.

And critically: `pq.read_table(store.get_object(key))` now succeeds against the fake (returned `SpooledTemporaryFile`, `seekable() → True`, reads 1000 rows from the synthetic Parquet in my repro). Same call path that hard-failed at the previous head_sha. Production parity is restored.

### Spool-spill tests — sound

`test_small_body_stays_in_memory_below_spool_threshold` (payload = `SPOOL_MAX_SIZE // 2`, asserts `_rolled is False`) and `test_large_body_spills_to_disk_above_spool_threshold` (payload = `SPOOL_MAX_SIZE + 4096`, asserts `_rolled is True`) cover the two practically-meaningful branches. I verified independently that `shutil.copyfileobj(body, spool, length=64*1024)` does in fact trip the rollover at the right threshold — small case stays `_rolled = False`, large case flips to `_rolled = True`, exactly-at-threshold stays in-memory (rollover triggers strictly above `max_size`).

The brief asked whether `_rolled` is the right inspection point given it's a private attribute. It is the practical target: CPython's `tempfile.SpooledTemporaryFile` has had `_rolled` since the original implementation, the class-level default is `_rolled = False` and `rollover()` sets it to `True` — there's no public API that exposes spill status, so the private attribute is the only signal. If a future Python version were to remove it the test would clearly fail on attribute access rather than silently pass, which is acceptable failure-mode for a test-only inspection.

### Arrow C++ pool delta assertion — measures steady-state, sufficient for the regression class

The `test_get_object_bounded_memory_for_multimb_parquet` test pairs `tracemalloc.get_traced_memory()` (Python pool peak) with `pa.default_memory_pool().bytes_allocated() - arrow_baseline` (Arrow C++ pool delta). Both assert `< 1.5x` payload size. Sayed asked for this; the pairing closes the gap where tracemalloc cannot see Arrow's C++ arena.

To answer the brief's question directly: the assertion measures **baseline-to-final**, not baseline-to-peak. The delta is computed after the spool's `with` block closes, while `table` is still bound, so `bytes_allocated()` reflects the steady-state working set of the live table — not any transient buffers that may have allocated and released during `pq.read_table`. A regression that allocated a separate full-size buffer during the read but released it before the table assignment would slip past this assertion.

That said — the regression class the assertion is built to catch (re-introducing a buffered Python `bytes` copy alongside Arrow's arena, the original pre-#12 shape) does persist its buffer until the read returns, so the Python-side `tracemalloc` peak DOES catch the most likely shape, and the Arrow-side delta catches the silent-arena-doubling case. The pairing is sufficient. A future hardening could use `pa.default_memory_pool().max_memory()` for a true peak rather than current allocation — that would catch transient-spike regressions too — but it's not blocking and Imelda would have to thread a `_total_allocated` reset across tests since `max_memory` is monotonic per-process. Out-of-scope for this PR; worth tracking as a hardening follow-up if anyone is curious.

### BytesIO-vs-spool comparison test — 5% tolerance is right

`test_spool_path_does_not_double_arrow_arena_vs_direct_bytesio` reads the same payload through `pq.read_table(io.BytesIO(payload))` (baseline) and `pq.read_table(spool)` (spool path), comparing the Arrow C++ delta with a 5% ceiling. Both paths feed identical bytes through the same pyarrow C++ readers, so the deltas should be near-identical and 5% leaves plenty of jitter room for pool-fragmentation noise. The brief asked whether 5% addresses my prior concern that 1% was potentially flaky — yes, fully. Good calibration.

### Fixture self-validation (`> 8 MiB`) — correctly aligned with `SPOOL_MAX_SIZE`

`_build_multimb_parquet(10 * 1024 * 1024)` builds a payload that compresses-none and uses no dictionary, so the on-disk size sits at ~1.0x the row bytes. With `row_bytes = 256` and `n_rows = 10 MiB / 256`, the actual Parquet payload comes out comfortably above 10 MiB. The `assert actual_payload_size > 8 * 1024 * 1024` floor explicitly aligns with `SPOOL_MAX_SIZE = 8 * 1024 * 1024` and guarantees the test exercises the spill path (not the in-memory branch). Good defensive coupling — if a future tweak shrinks the synthetic Parquet below the spool threshold, the test fails loudly instead of silently moving off the path it was designed to exercise.

### Consumer call-site wrapping — all 5 sites are `with`-wrapped

`dedup` (processor.py:241), `enrich` (processor.py:217), `normalize` (processor.py:476), `ingest` manifest read (processor.py:113), `ingest` node-parquet read (processor.py:379), `ingest` edges-parquet read (processor.py:393). Every `get_object` call site uses `with self.store.get_object(...) as stream:`, which properly closes the on-disk spool tempfile (if it spilled) deterministically. No leaked tmpfiles on disk in any code path. The `test_get_object_can_be_used_as_context_manager` test pins the contract — after `with`, `body.closed is True`.

### `serial` marker registration

`pyproject.toml` line ~120 now registers the marker alongside the existing `ml` and `integration` markers. Both memory tests are decorated with `@pytest.mark.serial`. With no `pytest-xdist` in `uv.lock` today this is purely defensive against future xdist adoption, but it's the right kind of defensive — explicit at the test site so a future `-n auto` doesn't quietly pollute the global pools.

### Out-of-scope hardening I noticed but am NOT blocking on

- `pa.default_memory_pool().max_memory()` would give a true peak rather than steady-state delta in `test_get_object_bounded_memory_for_multimb_parquet`. Catches transient-spike regressions the current assertion misses. Tracker only.
- `_rolled` is private; if CPython ever renames it the spool-spill tests would fail-loud on attribute access (acceptable). A future hardening would track it against the Python-version matrix in CI.
- Tests run serially in one process today (no xdist). The `serial` marker is registered defensively; not load-bearing yet.

### Overall

The PR now does what it advertises: bounded worker memory, real pyarrow compatibility, no fake-permissive gap between test and prod. Methodology is sound and the regression-class coverage is tight. Approving.

Requestor: Tomás Carvalho
Requestee: Imelda Santos
RequestOrReplied: Approved
TechDebt: none

Issue #43, opened 2026-05-17
bug(workers): WorkerRunner.handle_one mark-then-send ordering — downstream message lost on producer.send failure after mark


parametrization:
## Background

Surfaced during #11 / PR #41 review (Petra + Tomás, parallel reviewers). Pre-existing bug, NOT introduced by PR #41 — same failure mode existed on `InMemoryCheckpoint` before. Deferred from #41 scope per [[feedback_multi_layer_gap_filing]] / /file-bug Pass C — fixing it would expand #41 into runner-correctness territory beyond the durable-store gate.

## The race

`workers/lib/runner.py::WorkerRunner.handle_one` at HEAD `2f3ebdc`:

```python
# line 129
self.checkpoint.mark(msg.batch_id)
# line 132
self.producer.send(self.settings.produce_topic, serialize_message(next_msg))
```

If `producer.send` raises **after** `checkpoint.mark` returns, the batch_id is durably recorded as "processed" but the downstream message was never produced. On worker restart, `checkpoint.seen(batch_id) == True` → the message is silently skipped → permanent downstream loss.

The window is small (microseconds between `mark` commit and `producer.send` Future creation) but real. Failure modes that trigger it:
- Kafka broker unavailability between the two calls
- Producer queue full (`buffer.memory` exhausted)
- Process crash / OOM-kill in the window
- Network partition

## Why deferred from #41

- `#11` acceptance gate is "durable checkpoint across restart" — that gate IS satisfied; this bug is in the *use* of the checkpoint, not the checkpoint itself
- Fix touches `WorkerRunner.handle_one` — `#11` body explicitly carves out "no processor.py changes" but runner.py changes are also in expansion territory
- Pre-existing — `InMemoryCheckpoint` had the same ordering bug; #41 doesn't introduce it, just makes it more visible (since restarts now preserve state)

## Proposed fix shapes (pick one in scoping PR)

**Option A — send-then-mark with idempotency at consumer**: reverse the order. Downstream message may be re-produced after a mark-failure, but downstream consumers should already be idempotent via their own checkpoint scope. Risk: shifts the burden onto every downstream consumer's idempotency.

**Option B — transactional outbox**: stage the outbound message in a PG outbox table in the SAME transaction as the `mark`, then a separate dispatcher reads outbox → produces → marks-as-sent. Eliminates the race fully. Cost: adds a table + a dispatcher worker; more complex deployment.

[...]

Issue #11, opened 2026-04-21
tech-debt: replace InMemoryCheckpoint with durable idempotency store


parametrization:
## Context
PR #8 (Kafka workers scaffold) ships `InMemoryCheckpoint` in `workers/lib/runner.py` for `batch_id` idempotency. It loses all state on restart, so a worker crash + restart will re-process every in-flight pointer message it has already acked.

## Acceptable for
Wave-9 local dev / docker-compose smoke (single-process, short-lived).

## Required before
Any prod or shared-staging rollout. A restart in those environments would cause duplicate writes downstream (duplicate dedup decisions, duplicate enrichment, duplicate Neo4j MERGE work).

## Options to evaluate
- Postgres table keyed on `(stage, batch_id)` with TTL
- S3/B2 marker objects under a `checkpoints/{stage}/{batch_id}` prefix
- Kafka consumer-group offset commits paired with idempotent producer (`enable.idempotence=true`) + transactional output

## Acceptance
- Checkpoint survives worker restart in an integration test
- Swap is wired into all 4 workers without changing their `processor.py`

Tracked from PR #8 TechDebt attestation.

PR #48, opened 2026-05-31
fix(workers): send-before-mark in WorkerRunner.handle_one (Refs #43)

parametrization:
## Summary

`WorkerRunner.handle_one` marked the batch_id as processed (`checkpoint.mark`) **before** producing the downstream message (`producer.send`). If `producer.send` raised after the mark committed — broker unavailability, producer queue full, crash in the window — the batch was durably recorded as seen but its next-stage message was never produced. On restart the batch reads as seen and is silently skipped → permanent downstream loss.

## Fix (Option A — send-before-mark)

Reorder so the next-stage message is handed to the producer **first**, then `mark`. If `send` raises, the batch is not marked, so a restart reprocesses and re-sends it. A re-send is safe (downstream consumers absorb redelivery via their own idempotency scope); a lost message is not recoverable.

## Tests

Regression coverage for both checkpoint paths:
- `test_send_failure_after_mark_does_not_record_checkpoint_inmemory` — `InMemoryCheckpoint`
- `test_send_failure_after_mark_does_not_record_checkpoint_pg` — `PgCheckpoint` (via the unit pg-fake)
- `test_successful_send_then_marks_checkpoint` — happy path still marks

A producer that raises on the produce topic must leave the checkpoint unmarked and nothing landed downstream.

## Scope

Only `WorkerRunner.handle_one` ordering. No `processor.py` change, no `_Checkpoint` Protocol change (preserves #11's carve-outs).

Local: ruff check + ruff format --check + mypy src/ + full workers pytest all green.

Refs #43

### DIFF
--- a/tests/workers/test_runner.py
+++ b/tests/workers/test_runner.py
@@ -4,10 +4,32 @@ from __future__ import annotations
 
 import json
 
+import pytest
+
 from tests.workers.conftest import FakeConsumer, FakeProducer
+from tests.workers.test_checkpoint_pg import _FakePgConnection
+from workers.lib.checkpoint_pg import PgCheckpoint
 from workers.lib.dlq import DLQ_TOPIC
 from workers.lib.message import PipelineMessage, parse_message, serialize_message
-from workers.lib.runner import WorkerRunner, WorkerSettings
+from workers.lib.runner import InMemoryCheckpoint, WorkerRunner, WorkerSettings
+
+
+class _SendFailsProducer(FakeProducer):
+    """Producer whose downstream ``send`` raises, simulating broker
+    unavailability / queue-full in the mark/send window (#43).
+
+    DLQ sends still succeed so the failure is isolated to the
+    next-stage publish path.
+    """
+
+    def __init__(self, *, failing_topic: str) -> None:
+        super().__init__()
+        self._failing_topic = failing_topic
+
+    def send(self, topic: str, value: bytes) -> None:
+        if topic == self._failing_topic:
+            raise RuntimeError("broker unavailable")
+        super().send(topic, value)
 
 
 def _settings() -> WorkerSettings:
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Yusuke Inoue
Requestee: Léopold Mbongo
RequestOrReplied: Approved
TechDebt: none

Reviewed against issue #43 acceptance and the diff at head `1a734e56` (read via gh API; no checkout). CI green (lint-and-typecheck, security-audit, test all SUCCESS), branch MERGEABLE.

Verified all four acceptance items:

- Fix shape chosen — Option A (send-then-mark). The single `mark` in `handle_one` now runs after the downstream `producer.send`, and `send` raises out of the method (the try/except wraps only `self.process`), so on a send failure the batch is never marked. On restart it reprocesses and re-sends, which downstream consumers absorb via their own idempotency scope. The rationale comment is accurate to the implemented control flow.
- Implementation isolated to `workers/lib/runner.py::handle_one`; no `processor.py` touched and the `_Checkpoint` protocol surface is unchanged, both honoring the issue's out-of-scope carve-outs.
- The regression tests genuinely exercise the failure window. `_settings()` sets the produce topic, `process` returns a non-None next-stage pointer so the send branch is taken, and `_SendFailsProducer` raises only on the next-stage topic (DLQ sends still succeed), so the raise lands precisely in the post-process, pre-mark gap. Both tests assert the checkpoint was NOT recorded and nothing landed downstream — assertions that would fail under the old mark-then-send order, so they are true regression guards rather than vacuous.
- Both checkpoint backends covered — one test on `InMemoryCheckpoint`, one on `PgCheckpoint` via the in-memory `_FakePgConnection` fake — plus a happy-path test confirming a successful send still marks so redelivery is correctly skipped.

Clean, well-scoped, no concerns. Ship it.

parametrization:
Requestor: Petra Vidović
Requestee: Léopold Mbongo
RequestOrReplied: Approved
TechDebt: none

Approved. This is the clean send-then-mark fix (#43 Option A) and it lands exactly where the issue says it should — `WorkerRunner.handle_one`, no `processor.py` touched, no change to the `_Checkpoint` Protocol surface.

Correctness against the issue:
- The `checkpoint.mark` call now runs AFTER `producer.send`, so a `send` that raises propagates out of `handle_one` without recording the batch. On restart the batch reads as unseen and is reprocessed and re-sent — a re-send the downstream consumer absorbs via its own idempotency scope, which is the documented Option-A tradeoff and the right one for a streaming loader. No silent permanent loss.
- I confirmed there is no early-return between the send and the mark at HEAD, so the reorder cannot be short-circuited. The DLQ branch is upstream of both calls (process-failure path), so a process exception still DLQs and returns before either runs — unchanged and correct.

Test coverage is the strong part of this PR and directly satisfies the acceptance checklist:
- Both the `InMemoryCheckpoint` and `PgCheckpoint` paths get a dedicated regression that asserts `seen(...) is False` and `producer.sent == []` after the send raises — the issue explicitly required both paths covered.
- The `_SendFailsProducer` only fails the next-stage topic and lets DLQ sends through, so the failure window is isolated exactly as the race describes.
- The happy-path test pins the ordering from the other direction (send succeeds, then the batch is marked), which guards against a future refactor that drops the mark entirely.

CI is green on all three required checks (lint-and-typecheck, security-audit, test) and the branch is mergeable. No scope creep, no new tech debt.

Issue #38, opened 2026-05-17
test(ingest neo4j): real testcontainers integration coverage for wall-clock-bound contract


parametrization:
## Context

PR #36 (closes #20) removes a redundant outer `SessionExpired` reopen in `workers/ingest/processor.py`. The new contract relies on the neo4j driver's internal retry loop, bounded by `max_transaction_retry_time` (default 30s). The wall-clock bound itself is a **real-driver behavior** — the unit tests use `_FakeSession.execute_write` injection (`tests/workers/test_ingest_neo4j.py`) which bypasses `neo4j._sync.work.session._run_transaction` entirely, so the 30s bound contract documented in `processor.py`'s class docstring is **not exercised by any test in the repo**.

This is per [[feedback_test_mock_injection_masks_production_failure]] discipline: mock-bypass disclosures must point at a real tracking artifact for the bypassed contract.

## Problem

The PR documents a 30s per-message Neo4j retry budget as part of the public contract of `IngestProcessor.__call__`. That budget is enforced by:

- `neo4j._sync.work.session._run_transaction` retry loop (lines 494-586 in installed neo4j 6.1.0)
- `max_transaction_retry_time` config (`neo4j/_conf.py:360`, default 30.0s)

Neither surface is exercised by the unit suite. If a future driver upgrade silently changes the wall-clock semantics (e.g., default raised to 60s, or retry behavior changed under cluster mode), the documented "≤30s per message" contract becomes false and the unit tests do not catch it. The DLQ-on-budget-exhaustion behavior is also untested at the integration layer.

## Acceptance

Add a testcontainers-based test (`testcontainers[neo4j]>=4.0` already pinned in `pyproject.toml`) that:

1. **Wall-clock bound observation**: spin up a real Neo4j 5+ container, configure it to drop the connection mid-transaction (e.g., container restart, network partition via `pumba`, or `CALL dbms.cluster.routing.invalidate()` if cluster mode), assert that `IngestProcessor.__call__` returns/raises within `max_transaction_retry_time + tolerance` (target: ≤30s ± 5s with driver default).
2. **DLQ propagation under sustained failure**: with the broker bound exhausted, assert the runner-level DLQ topic receives the message with `error_class="SessionExpired"` (or whatever real-driver-final-error class the exhausted retry surfaces).
3. **Single-session invariant**: assert exactly one driver session is opened across the failure window (no compounding outer reopen).
[...]

Issue #33, opened 2026-05-17
ingest-worker: edge MERGE silently skips when endpoint MATCH returns empty — should fail loud per #22 acceptance


parametrization:
## Context

PR #18 + #21 landed the edge MERGE infrastructure (closed #22 2026-05-17). The "MERGE on node first" half of #22's acceptance is satisfied via manifest-ordered node-first execution + same-session writes. However, the "fail loud" half is not: `_build_edge_cypher` (`workers/ingest/processor.py:286`) uses:

```cypher
MATCH (s {id: row.src_id})
MATCH (t {id: row.dst_id})
MERGE (s)-[r:`<label>`]->(t)
```

If either endpoint node doesn't exist in the graph at MERGE time, the `MATCH` returns 0 rows and the `MERGE` of the relationship silently doesn't fire. No error, no DLQ, no metric. Within a single batch this is safe (the node MERGE runs first in the same session). But cross-batch — where an edge row references an endpoint that was supposed to land in a prior batch but never did — the edge data is lost silently.

The row-level validation in `_edge_rows_for_merge` (`workers/ingest/processor.py:231-235`) only catches missing-from-Parquet `src_id`/`dst_id`, not missing-from-graph.

## Problem

#22 acceptance literal: *"endpoint missing (both src_id and dst_id must exist — MERGE on node first or fail loud)"*. Today's behavior: silent skip, no telemetry.

This is a defensible Cypher idiom (skip-if-not-yet-ingested is a normal pattern for streaming graph loaders), but it diverges from #22's stated contract and creates an undetectable data-loss class for the never-landing-endpoint case.

## Severity / urgency

Low for now — ingest-platform is `status: planned`, no production traffic, no cross-batch edge volume yet. If silent edge skips occur during early bring-up, they'll be caught at first downstream graph-query audit. Becomes higher severity at first real ingest run.

## Proposed fix (pick one)

1. **`OPTIONAL MATCH` + null-check + DLQ**: rewrite `_build_edge_cypher` to use `OPTIONAL MATCH`, then `WITH row, s, t WHERE s IS NULL OR t IS NULL` -> emit a counter row that the runner DLQs. Preserves the Cypher single-statement shape but adds explicit "missing endpoint" telemetry.

[...]

Issue #22, opened 2026-04-22
tech-debt: ingest-worker — MERGE relationships (edges) after node MERGE lands


parametrization:
## Context

PR #18 (closes #13) implements node MERGE only. Relationships
(`TRANSMITTED_TO`, `NARRATED`, `GRADED_AS`, `IN_COLLECTION`,
`OCCURRED_DURING`, `LOCATED_IN`, etc. — see the canonical edge set in
`noorinalabs-main/ontology/domain.yaml`) are deferred.

This is a deliberate scope cut from #13 ("session per batch, MERGE on
stable ID") but the acceptance criteria in #13 also say "MERGEs
nodes/edges from the normalized Parquet" — so the edge half is still
owed.

## Acceptance

- `workers/ingest/processor.py` MERGEs edges in the same session (same
  retryable tx or a second tx in the same session)
- Edge types drawn from an allowlist mirroring `_ALLOWED_LABELS`
- Normalized Parquet carries an `edges` section (schema TBD — coordinate
  with #10) with `{src_id, dst_id, type, props}` rows
- Tests covering: edge allowlist rejection, happy-path MERGE, endpoint
  missing (both `src_id` and `dst_id` must exist — MERGE on node first
  or fail loud)
- Reference: `src/graph/load_edges.py` in noorinalabs-isnad-graph
  (batch size 1000, explicit SET per Phase 4 safety rule)

## Dependencies

- Blocked by: #18 (node MERGE), #10 (normalize must emit edges section)

parametrization:
Closed as done — bulk implementation landed via PR #18 + PR #21 (2026-04-22).

Evidence at wave-11 HEAD `5c3cadc`:
- `workers/ingest/processor.py:299` — `_merge_edges_tx` runs in same session as `_merge_nodes_tx`
- `workers/ingest/processor.py:277-289` — `_build_edge_cypher` with MATCH-then-MERGE
- `workers/ingest/schema.py:127-141` — `ALLOWED_EDGE_LABELS` frozenset (8 edge types: TRANSMITTED_TO, NARRATED, APPEARS_IN, STUDIED_UNDER, PARALLEL_OF, GRADED_BY, ACTIVE_DURING, BASED_IN)
- `workers/ingest/schema.py:147-186` — `EDGE_PROPERTY_MAP` per-edge property allow-lists
- `workers/ingest/processor.py:117-147` — `_split_manifest_entries` re-sorts node-first defensively

[...]

PR #49, opened 2026-05-31
fix(ingest): fail loud on edge MERGE with missing endpoint (Refs #33)

parametrization:
## Summary

Edge MERGE used `MATCH (s) MATCH (t) MERGE (s)-[r]->(t)`. When either endpoint did not exist at MERGE time, the `MATCH` returned 0 rows and the relationship MERGE silently never fired — no error, no DLQ, no metric. Cross-batch (an edge referencing an endpoint that should have landed in a prior batch but never did) this lost edge data silently — an undetectable data-loss class. It diverged from #22 acceptance: *"endpoint missing (both src_id and dst_id must exist — MERGE on node first or fail loud)"*.

## Change (issue option 1)

- `_build_edge_cypher`: `OPTIONAL MATCH` both endpoints; guard the relationship MERGE behind a `FOREACH` that fires only when both resolved; return both `merged` and `skipped` counts.
- `_merge_edges_tx`: raise new `EndpointMissingError` (subclass of `UnknownSchemaError` → runner's existing schema-error branch DLQs it) when `skipped > 0`; log `ingest_edge_endpoint_missing` with referenced endpoint ids for bring-up detection.
- Within-batch happy path unchanged (nodes MERGE first in the same session). Fail-loud keys on *graph presence*, not *this-batch presence*, so a legitimately prior-batch endpoint still merges cleanly.

## Acceptance (#33)

- [x] Missing endpoint = loud failure (EndpointMissingError → DLQ), not silent skip
- [x] Unit test: edge row referencing a node not present → schema error / DLQ
- [x] Docstring on `_build_edge_cypher` documents the chosen contract
- [x] Log line emitted on missing-endpoint event

## Tests

- OPTIONAL-MATCH / FOREACH / `skipped`-count Cypher shape
- missing endpoint → `EndpointMissingError` (message names the missing id)
- prior-batch endpoint present (graph-seeded) → clean merge, no error
- runner routes `EndpointMissingError` to DLQ with `error_class == "EndpointMissingError"`

Full unit suite: 587 passed, 4 skipped. 12 integration errors are pre-existing Docker/Neo4j-dependent (testcontainers, #38), unrelated. ruff check + format clean.

TechDebt: none

Refs #33
Refs #22

### DIFF
--- a/tests/workers/test_ingest_neo4j.py
+++ b/tests/workers/test_ingest_neo4j.py
@@ -17,6 +17,7 @@ import pytest
 
 from tests.workers.conftest import FakeConsumer, FakeProducer
 from workers.ingest.processor import (
+    EndpointMissingError,
     IngestProcessor,
     ManifestMissingError,
     UnknownSchemaError,
@@ -44,13 +45,47 @@ class _FakeResult:
 
 
 class _FakeTx:
-    def __init__(self, sink: list[tuple[str, dict[str, Any]]]) -> None:
+    """Records every ``tx.run`` and simulates the node/edge MERGE result shape.
+
+    Node MERGEs return ``{"merged": <row count>}``. Edge MERGEs use the
+    fail-loud ``OPTIONAL MATCH`` Cypher and return both ``merged`` and
+    ``skipped`` — the fake emulates the OPTIONAL MATCH by treating an
+    endpoint as present only if its id is in ``known_ids``. Node MERGEs
+    seen earlier in the same transaction add their ids to ``known_ids``,
+    mirroring the same-session "nodes first" ordering contract so the
+    happy path resolves every endpoint.
+    """
+
+    def __init__(
+        self,
+        sink: list[tuple[str, dict[str, Any]]],
+        *,
+        known_ids: set[str] | None = None,
+    ) -> None:
         self._sink = sink
+        # Endpoints resolvable by the simulated OPTIONAL MATCH. Seeded by
+        # the caller (cross-batch absent-endpoint tests) and grown by node
+        # MERGEs run earlier in this transaction (within-batch happy path).
+        self._known_ids: set[str] = set(known_ids or set())
 
     def run(self, query: str, **params: Any) -> _FakeResult:
         self._sink.append((query, params))
         rows = params.get("rows") or []
-        return _FakeResult({"merged": len(rows)})
+        is_edge = "MERGE (s)-[r:" in query
+        if not is_edge:
+            for row in rows:
+                row_id = row.get("id")
+                if isinstance(row_id, str):
+                    self._known_ids.add(row_id)
+            return _FakeResult({"merged": len(rows)})
+        merged = 0
+        skipped = 0
+        for row in rows:
+            if row.get("src_id") in self._known_ids and row.get("dst_id") in self._known_ids:
+                merged += 1
+            else:
+                skipped += 1
+        return _FakeResult({"merged": merged, "skipped": skipped})
 
 
 class _FakeSession:
@@ -59,9 +94,13 @@ class _FakeSession:
         sink: list[tuple[str, dict[str, Any]]],
         *,
         raise_on_execute: Exception | None = None,
+        known_ids: set[str] | None = None,
     ) -> None:
         self._sink = sink
         self._raise = raise_on_execute
+        # Pre-existing graph node ids the simulated OPTIONAL MATCH resolves
+        # for edge endpoints, on top of whatever nodes this batch MERGEs.
+        self._known_ids = known_ids
 
     def __enter__(self) -> _FakeSession:
         return self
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Camila Restrepo
Requestee: Imelda Santos
RequestOrReplied: Approved
TechDebt: none

Reviewed as data-lead (primary) against issue #33 acceptance and the #22 fail-loud contract. Read the diff and surrounding source at head sha 1d2429b (processor.py, runner.py, _edge_rows_for_merge, the batch closure) via the contents API — no checkout.

Fail-loud semantics are correct. `_build_edge_cypher` now `OPTIONAL MATCH`es both endpoints and guards the relationship MERGE behind a FOREACH that fires only when both sides resolve, returning per-row `merged` and `skipped` counts. `_merge_edges_tx` raises `EndpointMissingError` on any non-zero skip. The single-statement write-FOREACH-then-RETURN shape is valid, and because OPTIONAL MATCH preserves one row per unwound input even on a null endpoint, the per-row `count(CASE WHEN ...)` aggregation is cardinality-correct.

DLQ routing is sound. `EndpointMissingError` subclasses `UnknownSchemaError`, and the runner catches every exception, emitting `error_class=type(exc).__name__` to the DLQ topic — no new branch needed. The regression test asserts the record lands on DLQ_TOPIC with the expected class name.

Graph-presence vs this-batch-presence is distinguished correctly, which is the key data-contract point. The batch closure MERGEs all nodes first then edges in one transaction, so in-batch endpoints resolve; the prior-batch-endpoint test seeds the endpoint only in the graph (not this batch's node MERGEs) and shows the edge still merges with no raise. That proves a legitimate cross-batch edge is not falsely failed, while a never-landing endpoint fails loud.

Regression coverage is adequate: Cypher-shape assertion (OPTIONAL MATCH + FOREACH guard + both counts), in-batch missing endpoint raises with the missing id surfaced in the message, prior-batch-present merges cleanly, and end-to-end runner DLQ routing. The pre-existing nodes-then-edges happy-path test exercises in-batch endpoint resolution through the updated fake.

`_missing_endpoint_ids` honestly documents that it cannot pinpoint exactly which endpoints were absent without a second round-trip and surfaces the batch's distinct endpoint ids capped at 20 — a reasonable, bounded diagnostic for bring-up, not a correctness gap. No data-contract regression. CI green (lint-and-typecheck, security-audit, test all SUCCESS), mergeable. Approving.

parametrization:
Requestor: Petra Vidović
Requestee: Imelda Santos
RequestOrReplied: Approved
TechDebt: none

Approved. This implements the #33 fail-loud contract via Option 1 (OPTIONAL MATCH + null-count + DLQ), which is the option the issue recommended for behavioral parity with #22, and it does so without creating endpoint nodes (the rejected Option 3) so the `ALLOWED_NODE_LABELS` / Phase-4 safety gate is preserved.

Correctness against #22 acceptance ("both src_id and dst_id must exist — MERGE on node first or fail loud"):
- `_build_edge_cypher` now `OPTIONAL MATCH`es both endpoints and guards the relationship MERGE inside a `FOREACH (_ IN CASE WHEN s IS NOT NULL AND t IS NOT NULL THEN [1] ELSE [] END | ...)`, so the MERGE only fires when both sides resolved. A missing endpoint leaves the list empty — nothing is created — and the RETURN counts it via `count(CASE WHEN s IS NULL OR t IS NULL THEN 1 END) AS skipped`.
- `_merge_edges_tx` raises `EndpointMissingError` on any non-zero `skipped`, and that class subclasses `UnknownSchemaError`, so the runner's existing schema-error branch DLQs it with no new runner branch. I confirmed the subclass relationship in the diff and the test asserts it explicitly.
- The fail keys on *graph presence*, not *this-batch presence*: the cross-batch happy-path test (`test_edge_with_endpoint_in_prior_batch_present_merges_cleanly`) seeds the endpoint via the driver's `known_ids` and shows the edge MERGEs without raising — exactly the streaming-loader case #33 was worried about losing silently.

Acceptance checklist:
- Loud failure on missing endpoint — yes, `EndpointMissingError` raised, not silent skip.
- Unit test for the batch-missing case — `test_edge_with_missing_endpoint_in_batch_fails_loud`, asserting both the error type and that the missing id (`nar:b`) appears in the message.
- Runner DLQ routing — `test_runner_routes_missing_endpoint_to_dlq` asserts one DLQ record with `error_class == "EndpointMissingError"`.
- Docstring on `_build_edge_cypher` documents the chosen contract — present and clear.
- Metric/log on the missing-endpoint event — `_logger.error("ingest_edge_endpoint_missing", ...)` with edge_label, skipped, total, and the missing endpoint ids.

One detail I checked and like: `_missing_endpoint_ids` is honest about its limitation — the Cypher knows *that* an endpoint was missing but not *which* without a second round-trip, so it surfaces every endpoint id in the failing batch as the operator's starting point and caps at 20 to keep the DLQ record and log line bounded on a large batch. The cap carries a `# noqa: PLR2004` with a rationale that it's a log/message cap, not a domain constant — appropriate.

The test-fake updates (`_FakeTx` growing `known_ids` from node MERGEs seen earlier in the same tx, edge MERGEs returning `merged`/`skipped`) faithfully model the same-session nodes-first ordering, so the happy path resolves endpoints the way real Neo4j would within a batch.

CI green on all three checks, mergeable. No scope creep, no new tech debt.

Issue #20, opened 2026-04-22
ingest-worker: bound total Neo4j retry time — SessionExpired layer compounds with driver retry


parametrization:
## Context

PR #18 layers retry logic for Neo4j writes:

- **Inner**: `session.execute_write` uses neo4j driver's built-in retry for transient errors (`ServiceUnavailable`, `TransientError`). Default `max_transaction_retry_time` is **30s**.
- **Outer**: `IngestProcessor.__call__` catches `SessionExpired` and reopens + re-runs the whole grouped-rows loop once.

## Problem

Under a real outage, the two layers compound:

1. Driver retries `ServiceUnavailable` for up to 30s → raises.
2. If the raised exception is `SessionExpired`, outer handler opens a new session and runs the loop again — another 30s retry window per label batch.

**Worst-case latency per message**: ~60s (or 30s × N labels if SessionExpired hits mid-loop). No total timeout bound. Under sustained Neo4j degradation this stalls the consumer lag rather than shedding load to DLQ where ops can triage.

## Proposed fix

One of:

1. **Wall-clock bound**: wrap the outer retry in a deadline (`time.monotonic()` check) — e.g., 45s total budget, then DLQ.
2. **Tune driver**: lower `max_transaction_retry_time` on the driver to 15s so 2× compound stays under 30s.
3. **Drop the outer SessionExpired retry**: driver 5.x already retries `SessionExpired` inside `execute_write` for routing-table refreshes; the manual reopen is likely redundant. Verify and remove.

Option 3 is lowest-risk if verified — the manual reopen was written defensively but may be unnecessary with current driver semantics.

## Acceptance

- [ ] Total Neo4j retry time per message has a documented upper bound
- [ ] Unit test: sustained transient errors DLQ within the bound rather than blocking the consumer indefinitely
- [ ] Verify whether the manual `SessionExpired` reopen is redundant with neo4j-driver ≥5.15

Refs #13, #18.

PR #52, opened 2026-05-31
test(ingest neo4j): testcontainers integration coverage (Refs #38)

parametrization:
## Summary

Adds `tests/integration/test_ingest_neo4j_wall_clock.py` — real-Neo4j (testcontainers) integration coverage for the per-message retry-budget contract documented in `IngestProcessor.__call__` but **not exercised by any unit test**. The unit suite's `_FakeSession.execute_write` raises directly and bypasses `neo4j._sync.work.session._run_transaction` entirely, so the `max_transaction_retry_time` wall-clock bound is a real-driver behavior with zero coverage (mock-bypass disclosure per the `test_mock_injection_masks_production_failure` discipline this issue cites).

## Coverage (Refs #38 acceptance)

1. **Wall-clock bound observation (item 1)** — `test_wall_clock_bound_is_enforced_by_real_driver`: seed a one-node batch, kill the container *before* the processor opens its session, configure a short `max_transaction_retry_time`, and let the driver's internal retry loop spin against the dead server. Asserts the processor raises within `budget + tolerance` — a regression removing the bound would hang far longer. Uses a short budget so the same `_run_transaction` mechanism the 30s default uses is observed without a 30s wait.
2. **DLQ propagation under sustained failure (item 2)** — `test_exhausted_retry_routes_to_dlq_via_runner`: route the exhausted-retry batch through `WorkerRunner.handle_one`; assert exactly one DLQ record with a real transient driver `error_class` (`ServiceUnavailable`/`SessionExpired`/`TransientError`).
3. **Single-session invariant (item 3)** — a `_CountingDriver` proxy over the real driver asserts exactly one session is opened across the failure window (no compounding #20 outer reopen). Re-asserted on the success path.

Plus a happy-path control (`test_real_driver_merges_node_when_container_is_live`) that MERGEs against a live container and independently reads the node back — anchoring that the failure-path tests fail because the container is *killed*, not because the wiring is broken.

**Item 4 (cluster-mode fresh-acquire)** requires a real Enterprise causal cluster the single-instance community testcontainer cannot provide. Documented in the module docstring as a follow-up; the single-instance observable proxy (one session, fresh connection per retry iteration) is asserted instead.

## Gating / runtime

- Marked `@pytest.mark.integration` (the marker already registered in `pyproject.toml`); needs Docker for testcontainers.
- **Docker-gated, not verified live in this environment** — Docker is not running here, so the container suite was not executed. Verified instead: the module **collects cleanly** (3 tests), is **deselected** by `-m "not integration"` (the exact filter CI's `ci.yml` and `make check` use), and the full non-integration suite stays green (`581 passed, 4 skipped, 18 deselected`). The unit ingest suite (`20 passed`) is unaffected. Runs via `make test-integration` where Docker is present — same convention as the existing `tests/integration/` suite. `importorskip("testcontainers.neo4j")` keeps a Docker-less lane collecting cleanly.
- `ruff check` + `ruff format --check` clean on both touched files.

Also points the unit suite's `#38` testcontainers references at the concrete new module.

**TechDebt:** none

Refs #38

### DIFF
--- a/tests/integration/test_ingest_neo4j_wall_clock.py
+++ b/tests/integration/test_ingest_neo4j_wall_clock.py
@@ -1,0 +1,338 @@
+"""Real-Neo4j integration coverage for the ingest wall-clock retry bound (#38).
+
+``IngestProcessor.__call__`` documents a per-message Neo4j retry budget as
+part of its public contract: transient driver errors
+(``ServiceUnavailable``, ``TransientError``, ``SessionExpired``) are retried
+*inside* ``session.execute_write`` by the neo4j driver's internal
+``_run_transaction`` loop, bounded by ``max_transaction_retry_time``
+(default 30s). Once that wall-clock budget is exhausted the last error
+propagates and the runner DLQs the message.
+
+The unit suite (``tests/workers/test_ingest_neo4j.py``) exercises the *wire
+contract* — one session opened, exception propagates, runner DLQs — but its
+``_FakeSession.execute_write`` raises directly and therefore bypasses
+``neo4j._sync.work.session._run_transaction`` entirely. The wall-clock bound
+itself is a **real-driver behavior** that no mock can stand in for: if a
+future driver upgrade silently changes the retry semantics (default raised,
+retry behavior changed under cluster mode) the documented "<=30s per
+message" contract becomes false and the unit tests do not catch it.
+
+Per [[feedback_test_mock_injection_masks_production_failure]], the
+mock-bypass disclosure in the unit suite points here. This module exercises
+the bound against a real Neo4j container by:
+
+* driving the driver's internal retry loop against a server that has been
+  killed mid-flight (``container.stop()``), so the loop runs until the
+  wall-clock budget is exhausted rather than succeeding on the first try;
+* configuring a *short* ``max_transaction_retry_time`` so the test observes
+  the same mechanism the 30s default uses without waiting 30s;
+* asserting the processor returns/raises within ``budget + tolerance``;
+* asserting exactly one driver session is opened across the failure window
+  (the [[#20]] single-session invariant — no compounding outer reopen);
+* asserting the runner routes the exhausted-retry failure to the DLQ with a
+  real-driver-final error class.
+
+Runtime requirement
+-------------------
+These tests are marked ``@pytest.mark.integration`` and need Docker on the
+host (testcontainers spins up ``neo4j:5-community``). They are excluded from
+the default ``make check`` / ``make test-workers`` run (``-m "not
+integration"``) and run via ``make test-integration`` where Docker is
+available (CI integration job / local dev with Docker). Without Docker the
+whole module is collected-but-skipped cleanly — see the module-level
+``importorskip`` gate below.
+
+Cluster-mode caveat (acceptance item #4)
+----------------------------------------
+The "fresh-acquire lands on a different cluster member" assertion from the
+issue requires a real Neo4j *cluster* (Enterprise causal cluster), which the
+single-instance community testcontainer cannot provide. We assert the
+observable single-instance proxy instead — exactly one session opened and a
+fresh connection acquired per retry iteration — and leave the multi-member
+routing assertion as a documented follow-up for an Enterprise-licensed
+cluster harness (not exercisable on the community image).
+"""
+
+from __future__ import annotations
+
+import json
+import time
+from typing import Any
+
+import pytest
+
+# testcontainers + a Docker daemon are hard requirements; without either the
+# module is collected-but-skipped so a Docker-less CI lane (make check) stays
+# green. importorskip on the package is the same gate the PG-restart
+# integration test relies on indirectly via the session fixtures.
+pytest.importorskip("testcontainers.neo4j")
+
+from neo4j import Driver, GraphDatabase  # noqa: E402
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Imelda Santos
Requestee: Tomás Carvalho
RequestOrReplied: Approved
TechDebt: none

## Review — PR #52 (Refs #38), verified at head 0cc97b6

Read AC on #38, the full diff at head via gh API (no checkout), CI, and the referenced fixtures/markers/Makefile/CI workflow.

### Gating — verified clean
- New module sets `pytestmark = pytest.mark.integration`. The `integration` marker IS registered in `pyproject.toml` (markers list), so no unknown-marker warning.
- CI test step (`.github/workflows/ci.yml`) runs `uv run pytest tests/ -m "not integration"`, so the whole module is deselected on the Docker-less CI lane — never collected, never needs Docker. `make test-integration` (`-m integration`) runs it where Docker exists. Gating is correct in both directions.
- Module-level `pytest.importorskip("testcontainers.neo4j")` is a harmless belt-and-suspenders fallback (testcontainers is a hard dep, so it rarely fires; the marker is the real gate).
- Function-scoped `neo4j_container_killable` correctly avoids killing the session-scoped `neo4j_container` shared by the rest of `tests/integration/`. Good catch by the author.

### Contract coverage — genuine, not trivial
- `test_wall_clock_bound_is_enforced_by_real_driver` drives the real driver's `_run_transaction` retry loop against a killed container with `max_transaction_retry_time=5s` and asserts raise-within-budget+tolerance plus the [[#20]] single-session invariant via the `_CountingDriver` proxy. This exercises the actual bound the unit suite's `_FakeSession.execute_write` injection bypasses — exactly the mock-bypass gap #38 was filed to close.
- DLQ-propagation test asserts the runner routes the exhausted-retry failure to the DLQ with a real transient error class (membership in the documented transient family rather than a brittle exact pin — appropriate given driver-version variance).
- Happy-path control anchors the failure tests (proves failures come from the kill, not broken wiring) and reads back the MERGE'd node via an independent session.
- Upper-bound-only timing assertion is the right call — a lower bound on connection-refused timing would be flaky, and the upper bound is what guards the "bound removed → unbounded hang" regression per AC item 1.

### Scope honesty
AC item 4 (fresh-acquire lands on a different cluster member) requires an Enterprise causal cluster the community testcontainer cannot provide. The PR asserts the single-instance observable proxy (one session + fresh connection per retry) and documents the multi-member routing assertion as an Enterprise-cluster follow-up. Reasonable and clearly disclosed rather than faked.

CI: lint-and-typecheck / security-audit / test all SUCCESS. mergeable MERGEABLE.

Approved.

parametrization:
Requestor: Petra Vidović
Requestee: Tomás Carvalho
RequestOrReplied: Approved
TechDebt: none

Approved. This adds the real-Neo4j testcontainers coverage #38 asks for, exercising the wall-clock retry bound that `_FakeSession.execute_write` injection cannot reach, and it does the infra wiring correctly so the Docker-less CI lane stays green.

Acceptance walk-through:
- Item 1 (wall-clock bound): `test_wall_clock_bound_is_enforced_by_real_driver` kills the container before the session opens, drives the driver's internal `_run_transaction` loop to budget exhaustion, and asserts the processor raises within `RETRY_BUDGET_S + TOLERANCE_S`. Using a short `max_transaction_retry_time` to observe the same mechanism the 30s default uses — rather than waiting 30s — is the right call and keeps the suite fast. A regression that removed the bound would hang far past the assertion.
- Item 2 (DLQ on sustained failure): `test_exhausted_retry_routes_to_dlq_via_runner` runs the killed-container batch through the real `WorkerRunner`, asserts exactly one DLQ record on `DLQ_TOPIC`, and checks `error_class` is one of the retryable transient family the processor docstring enumerates rather than pinning the exact class — appropriately tolerant of driver-version/failure-mode variation.
- Item 3 (single-session invariant): the `_CountingDriver` proxy wraps a real driver and asserts `session_count == 1` across the failure window in both failure tests and the live-container control — directly proving no compounding outer reopen (the #20 contract).
- Item 4 (cluster fresh-acquire on a different member): honestly scoped out. The single-instance community testcontainer cannot host a causal cluster, so the PR asserts the observable single-instance proxy and documents the multi-member routing assertion as a follow-up needing an Enterprise cluster harness. That is the correct call under runtime-gate-scoping rather than faking a cluster.

Things I checked beyond the diff:
- The `integration` marker is registered in `pyproject.toml`, the CI `test` job runs `pytest -m "not integration"` (ci.yml), and the module has a `pytest.importorskip("testcontainers.neo4j")` gate plus `pytestmark = pytest.mark.integration`. So green CI is a meaningful signal here — the new file is collected-but-skipped in the Docker-less lane, not silently passing.
- `make test-integration` runs `tests/integration/ -m integration`, and the referenced session-scoped `neo4j_container` fixture exists in `tests/integration/conftest.py`; the function-scoped `neo4j_container_killable` is correctly separate because the wall-clock test stops the container.
- The happy-path control (`test_real_driver_merges_node_when_container_is_live`) anchors the failure tests by proving the wiring works against a live container and the node actually lands via an independent read session.

The two-line docstring update in `tests/workers/test_ingest_neo4j.py` repoints the old "tracked separately in issue #38" disclosure at the concrete integration module, which closes the mock-bypass-disclosure loop per the discipline #38 cites.

CI green on all three checks, mergeable. No scope creep, no new tech debt.

Issue #35, opened 2026-05-17
rationalize Phase-3-legacy ingest fields against Phase-4 Pydantic models


parametrization:
## Context

Filed pre-emptively during noorinalabs/noorinalabs-isnad-ingest-platform#24 codegen authoring — fully understands the asymmetry, just wants a GH home for the rationalization decision so it doesn't get lost when #24 closes.

The #24 codegen mechanism introduces (in `noorinalabs-isnad-graph`):
- `scripts/emit_ingest_schema.py` — derives the canonical NODE_PROPERTY_MAP from `src/models/*` at build time
- `.github/workflows/schema-drift.yml` — CI gate that hard-fails on un-rationalized divergence in either direction
- `scripts/ingest-extras.yaml` — declares ingest-allowed property keys that aren't on the Pydantic models, with one-line rationale + a followup-issue ref per entry

This issue is the **default rationalization target** that `ingest-extras.yaml` entries point at when their rationale category is `phase-3-legacy`.

## The bidirectional drift (HEAD evidence at origin/deployments/phase-3/wave-11)

Verified against `noorinalabs-isnad-graph/src/models/*.py` and `noorinalabs-isnad-ingest-platform/workers/ingest/schema.py`.

### Hadith — 8 ingest-only fields

In `NODE_PROPERTY_MAP["Hadith"]` but NOT on `Hadith` Pydantic model:
- `grade`
- `sect`
- `collection_name`
- `book_number`
- `chapter_number`
- `hadith_number`
- `chapter_name_ar`
- `chapter_name_en`

### Narrator — 2 ingest-only fields

In `NODE_PROPERTY_MAP["Narrator"]` but NOT on `Narrator` Pydantic model:
- `name_ar_normalized`
- `transmission_method`

### Collection — 1 ingest-only field

In `NODE_PROPERTY_MAP["Collection"]` but NOT on `Collection` Pydantic model:
- `source_corpus`

### Why this drift exists

`workers/ingest/schema.py` (PR #18, D-ii rewire) and `src/models/*.py` (commit d634085, "Phase 0") were authored independently across two repos and never co-evolved. The drift isn't a recent regression — it's the steady state of two hand-authored mirrors. #24 mechanizes detection going forward; this issue captures the cleanup decision for what already exists.

## The data-team decision

For each ingest-only field above, the data team (Farhan + Adaeze) decides one of:

[...]

PR #54, opened 2026-05-31
refactor(ingest): rationalize per-appearance Hadith fields per #35 ruling (Refs #35)

parametrization:
## Summary

Lands the **ingest-platform-side** parts of the ratified #35 data-team ruling that do NOT depend on Farhan's isnad-graph model PR. (The four PROMOTE fields' `ingest-extras.yaml` drops wait on his model change — out of scope here.)

### REMOVE node-copy — `book_number` / `chapter_number` / `hadith_number`
Per-appearance facts the model carries on the `APPEARS_IN` edge, not Hadith-node properties.

**Owner-gated query-layer check (PASSED for all three):** verified org-wide that no query/API/frontend consumer reads these off the Hadith **node**:
- isnad-graph `src/api`, `queries/` (Cypher), `frontend/src` — zero references
- user-service / landing-page / deploy `src/` — zero references
- (data-acquisition `src/graph/load_nodes.py` still *writes* `n.book_number` etc. via a separate legacy Phase-3 loader, but nothing **reads** them back off the node — a writer, not a query-layer reader. Gate passes. Flagged separately to TPM as legacy-writer context.)

Removed from `NODE_PROPERTY_MAP["Hadith"]` and from normalize's Hadith-node payload (so no per-batch "dropped unknown prop" warning). The `APPEARS_IN` edge keeps all three.

### EDGE canonicalization — `APPEARS_IN.hadith_number` → `hadith_number_in_book`
The `AppearsIn` Pydantic model field is `hadith_number_in_book`; the bare legacy `hadith_number` is dropped from `EDGE_PROPERTY_MAP["APPEARS_IN"]` and normalize emits the value under the modeled key. Source column stays `hadith_number` (feeds the edge).

### Not in this PR (deliberate)
- **PROMOTE** (`chapter_name_ar`, `chapter_name_en`, `name_ar_normalized`, `source_corpus`): wait on Farhan's isnad-graph model PR; their ingest-extras drops land after.
- **KEEP→permanent** flips (`grade`, `sect`, `collection_name`, `transmission_method`) and the dead-entry drops for the removed fields: these edit `noorinalabs-isnad-graph/scripts/ingest-extras.yaml` — an **isnad-graph** file, so a separate isnad-graph PR (cross-repo boundary). The schema-drift gate stays green either way (verified).

## Verification
- `scripts/emit_ingest_schema.py check` against the modified ingest schema → **clean** (gate not broken).
- Regression test: Hadith node omits the three fields; `APPEARS_IN` carries `hadith_number_in_book` not bare `hadith_number`.
- Full non-integration suite: 596 passed, 4 skipped. ruff check + format clean.

TechDebt: none

Refs #35

### DIFF
--- a/tests/workers/test_processors.py
+++ b/tests/workers/test_processors.py
@@ -488,6 +488,52 @@ class TestNormalizeProcessor:
         assert len(hadiths) == 1
         assert hadiths[0]["props"]["collection_name"] == "bukhari"
 
+    def test_hadith_node_omits_per_appearance_fields_per_issue_35(
+        self,
+        object_store: ObjectStore,
+        sample_message: PipelineMessage,
+        hadith_row: Any,
+    ) -> None:
+        """#35 ruling: book/chapter/hadith number are per-appearance facts.
+
+        They are no longer emitted onto the Hadith node (removed from the
+        ingest NODE allow-list; no query/API/frontend consumer reads them
+        off the node). They live on the APPEARS_IN edge instead, and the
+        edge carries the modeled name ``hadith_number_in_book`` rather than
+        the bare legacy ``hadith_number``.
+        """
+        row = hadith_row(source_id="h1", matn_en="text", collection_name="bukhari")
+        row["book_number"] = 3
+        row["chapter_number"] = 7
+        row["hadith_number"] = 42
+        _seed(object_store, sample_message.b2_path, build_hadith_parquet([row]))
+
+        nxt = NormalizeProcessor(object_store)(sample_message)
+
+        # Node side: none of the three per-appearance fields on the Hadith node.
+        hadiths = self._read_nodes(object_store, nxt.b2_path, "hadiths.parquet")
+        assert len(hadiths) == 1
+        node_props = hadiths[0]["props"]
+        for removed in ("book_number", "chapter_number", "hadith_number"):
+            assert removed not in node_props, f"{removed} must not be on the Hadith node"
+
+        # Edge side: APPEARS_IN keeps book/chapter and uses the canonical
+        # ``hadith_number_in_book`` key, not the bare ``hadith_number``.
+        edges_table = pq.read_table(object_store.get_object(f"{nxt.b2_path}edges.parquet"))
+        edge_rows = edges_table.to_pylist()
+        appears = [
+            {**e, "props": json.loads(e["props"])} for e in edge_rows if e["label"] == "APPEARS_IN"
+        ]
+        assert len(appears) == 1
+        edge_props = appears[0]["props"]
+        assert edge_props.get("book_number") == 3
+        assert edge_props.get("chapter_number") == 7
+        assert edge_props.get("hadith_number_in_book") == 42
+        assert "hadith_number" not in edge_props, (
+            "APPEARS_IN must use the canonical hadith_number_in_book, "
+            "not the bare legacy hadith_number"
+        )
+
     def test_missing_required_column_raises(
         self,
         object_store: ObjectStore,
--- a/workers/ingest/schema.py
+++ b/workers/ingest/schema.py
@@ -81,9 +81,12 @@ NODE_PROPERTY_MAP: dict[str, list[str]] = {
         "source_corpus",
         "sect",
         "collection_name",
-        "book_number",
-        "chapter_number",
-        "hadith_number",
+        # book_number / chapter_number / hadith_number removed from the
+        # Hadith node allow-list per the #35 data-team ruling: these are
+        # per-appearance facts the model carries on the APPEARS_IN edge,
+        # never read off the Hadith node by any query/API/frontend
+        # consumer (verified org-wide at #35 execution time). The edge
+        # keeps them (see EDGE_PROPERTY_MAP["APPEARS_IN"]).
         "chapter_name_ar",
         "chapter_name_en",
     ],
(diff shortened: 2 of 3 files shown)


parametrization:
Requestor: Camila Restrepo
Requestee: Imelda Santos
RequestOrReplied: Approved
TechDebt: none

Reviewed as data-lead (primary) against the owner-ratified #35 ruling. Read the diff and the full schema.py + normalize/processor.py at head sha bd1b734 via the contents API — no checkout.

Node-side removal is correctly scoped to the Hadith NODE payload only. `NODE_PROPERTY_MAP["Hadith"]` drops book_number / chapter_number / hadith_number (chapter_name_ar / chapter_name_en stay), and the normalize node block stops emitting them, which also avoids the per-batch dropped-unknown-prop warning. Nothing else in schema.py or the normalize processor still reads those three off the node.

The APPEARS_IN edge keeps the per-appearance facts. `EDGE_PROPERTY_MAP["APPEARS_IN"]` retains book_number and chapter_number and now carries hadith_number_in_book; the bare legacy hadith_number is dropped from both the edge allow-list and the normalize edge payload. That matches the AppearsIn Pydantic model field name, so this is a true canonicalization, not a data drop.

Source columns are untouched. Normalize still reads row.get for book_number, chapter_number, and hadith_number from the input row to populate the edge; the input hadith_number column stays as-is and only the emitted edge key is renamed. No required-column or input-contract change, so no upstream regression.

Regression coverage asserts both sides in one test. The new test confirms the three fields are absent from the Hadith node props and that the APPEARS_IN edge carries book_number, chapter_number, and hadith_number_in_book with the matching seeded values while the bare hadith_number is absent from the edge props. Node-absence and edge-canonicalization are both pinned.

No node-readers org-wide (Imelda's query-layer audit, and the one legacy writer filed as data-acq#59 is non-blocking). CI green (lint-and-typecheck, security-audit, test all SUCCESS), mergeable. No data-contract regression. Approving.

parametrization:
Requestor: Petra Vidović
Requestee: Imelda Santos
RequestOrReplied: Approved
TechDebt: none

Approved. This lands exactly the ingest-platform-side slice of the #35 ruling that doesn't depend on Farhan's isnad-graph model PR, and the node/edge payload changes match the ruling one-for-one.

I verified the ruling-match against the actual diff and the files at HEAD rather than the PR body framing:

Node side (REMOVE the three per-appearance fields):
- `NODE_PROPERTY_MAP["Hadith"]` at HEAD no longer lists `book_number` / `chapter_number` / `hadith_number` — only the explanatory comment remains, and the surrounding node fields (`source_corpus`, `sect`, `collection_name`, `chapter_name_ar/en`) are untouched.
- normalize's Hadith-node payload stops emitting all three, which is the right move — leaving them emitted but stripped at the allow-list would spray a per-batch dropped-unknown-prop warning. The node payload otherwise unchanged.

Edge side (CANONICALIZE to the modeled name):
- `EDGE_PROPERTY_MAP["APPEARS_IN"]` keeps `book_number` and `chapter_number`, drops the bare legacy `hadith_number`, and adds `hadith_number_in_book`.
- I confirmed against `noorinalabs-isnad-graph/src/models/edges.py` at wave-13 HEAD that the `AppearsIn` model field is `hadith_number_in_book` (and the bare `hadith_number` is not on the model), so the new edge key is the genuinely model-canonical one.
- normalize emits the value under `hadith_number_in_book` while still reading the input as `row.get("hadith_number")` — i.e. the source/input column stays `hadith_number` and only the emitted edge key changed. That matches the ruling's "source column unchanged" exactly.

Schema-drift gate stays clean — and I checked this cross-repo rather than trusting the ingest-repo CI alone, since `scripts/emit_ingest_schema.py check` runs in isnad-graph and fetches the live ingest `schema.py`, so this repo's green CI does not itself exercise that gate:
- The gate's `_check_drift` only errors on (a) a model field absent from ingest's allow-list, or (b) an ingest allow-list field that is neither on the model nor in `ingest-extras.yaml`. The three removed Hadith fields are not on the model, so dropping them from `ingest_set` cannot trip direction (a), and they simply leave the phantom direction too.
- `hadith_number_in_book` now sits in both `ingest_set` and the `AppearsIn` model set, so it is excluded from the phantom direction — no extras entry even required for it.
- The `ingest-extras.yaml` entries for the removed fields become dead, but the gate never requires an extras entry to correspond to a live ingest field (it only subtracts `extras_set` from `ingest_set`), so the orphaned entries are harmless and the gate stays green. Cleaning them up correctly belongs to the separate isnad-graph PR the PR body defers, alongside the KEEP-permanent flips and the PROMOTE drops — that cross-repo boundary call is right; none of it can be done from this repo.

Scope discipline is good: the four PROMOTE fields and the KEEP-permanent flips are explicitly held for the isnad-graph model PR / a separate isnad-graph PR, so this PR touches only ingest-platform files and only the REMOVE + canonicalize decisions, which is the subset that is self-contained here.

The regression test asserts both halves of the ruling at the payload layer — the Hadith node omits all three fields, and the APPEARS_IN edge carries `hadith_number_in_book == 42` (the seeded value) while explicitly asserting `hadith_number` is absent from the edge props. That pins the canonicalization from both directions.

CI green on all three checks (lint-and-typecheck, security-audit, test), mergeable. No scope creep, no new tech debt.

PR #73, opened 2026-06-12
feat(api): admin-only HTTP reset endpoints wrapping the reset CLI surface (#70)

### LINKED ISSUE #70: feat: HTTP reset endpoint exposing the reset CLI surface (for admin UI; split from main#138)
## Context

Split out from **noorinalabs-main#138** at P4W3 kickoff (pre-flight 0.3). The admin pipeline-reset **UI** lives in isnad-graph; the **HTTP endpoint** that exposes the reset CLI surface belongs here, in the ingest-platform repo.

## Scope (this repo — ingest-platform)
- HTTP endpoint(s) wrapping the existing reset CLI surface (ingest-platform#9, merged):
  - stage reset, source reset, full reset/obliterate
- Auth: admin-only (mirror the isnad-graph admin restriction; reachable from the admin dashboard backend)
- Preserve the CLI's confirmation/obliterate semantics at the API boundary
- Audit-log emission per reset call

## Consumers
- isnad-graph admin reset UI (relocated companion issue) calls this endpoint.

## Provenance
New issue carved from noorinalabs-main#138 (closed, relocated) per the wave-kickoff pre-flight 0.3 relocation rule. Builds on the reset CLI from ingest-platform#9.

parametrization:
## Summary

Exposes the pipeline **reset CLI surface** (#9) over HTTP so the isnad-graph admin reset UI (ig#970) can drive pipeline resets. Carved from noorinalabs-main#138 at P4W3 kickoff.

Three admin-only endpoints under `/admin/reset`, each routing through the **real** `PipelineResetter` and emitting an audit entry per call:

| Endpoint | Scope | Body |
|---|---|---|
| `POST /admin/reset/stage`  | one stage's B2 prefix + Kafka offsets | `{stage, dry_run?}` |
| `POST /admin/reset/source` | every stage prefix for one source | `{source, dry_run?}` |
| `POST /admin/reset/full`   | full obliterate (B2 + Kafka + Neo4j + PG) | `{confirmation, dry_run?}` |

## Contract decisions

- **Admin-only** — mirrors the isnad-graph admin restriction: the *same* user-service RS256 JWT (validated against user-service JWKS, role resolved from the `roles` claim) authorizes here. `require_admin` is applied at the router mount, so every reset route is guarded. Non-admin → **403**, missing/invalid/non-`access` token → **401**, JWKS unreachable → **503**. `src/api/auth.py` is a deliberately slimmed copy of isnad-graph's auth path (same role map, JWKS TTL cache, single key-rotation retry).
- **OBLITERATE confirmation preserved at the API boundary** — `/full` rejects any request whose `confirmation != "OBLITERATE"` (**400**) *before* touching the reset surface; a missing field is **422**. The supplied token maps to the CLI's typed-confirm path, recorded as `confirmation_method = "interactive"` for SIEM parity.
- **Audit per call** — real resets via `PipelineResetter.reset`, dry-runs via `write_dry_run_audit` (both emit an entry), matching the CLI exactly.
- **One client builder** — reuses `cli._build_reset_clients`; resetter + data dir are FastAPI deps so the unit suite runs with no boto3/Kafka/Neo4j/PG.

## Endpoint contract for ig#970 (admin UI)

Base path `/admin/reset`. Send `Authorization: Bearer <admin user-service JWT>`. Full reset requires `{"confirmation": "OBLITERATE"}`. All three accept optional `dry_run: true` (validates scope + writes a dry-run audit entry, touches no infra). Response: `{level, dry_run, confirmation_method, audit_entry_path, summary}`. App served via `uvicorn src.api.app:create_app --factory`; unauthenticated `GET /healthz` liveness probe.

## Testing

- `tests/test_api/test_reset_endpoints.py` — 12 tests across the three bars: auth gate (401 / 403 / admin-through / non-access-type), confirmation-required (full reset rejected before the surface is touched), and each endpoint driving the resetter with the correct scope; dry-run writes an audit entry without invoking the live resetter.
- CI-equivalent local run: `mypy src/` clean, `ruff check`/`format --check` clean over `src workers tests`, full suite **617 passed, 3 skipped** (non-integration/non-ml).
- **Live leg** (real MinIO/B2 + Kafka + Neo4j + PG wipe) needs the running stack — covered by the existing reset suite + CLI path. This PR adds and tests the **HTTP layer**; that contract correctness is the PR-time bar.

Closes #70

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,9 @@ dependencies = [
     "rapidfuzz>=3.6",
     "kafka-python>=2.0",
     "boto3>=1.34",
+    "fastapi>=0.115",
+    "uvicorn[standard]>=0.34",
+    "python-jose[cryptography]>=3.3",
 ]
 
 [project.scripts]
@@ -106,6 +109,10 @@ ignore_missing_imports = true
 
 [[tool.mypy.overrides]]
 module = ["kaggle", "kaggle.*"]
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = ["jose", "jose.*"]
 ignore_missing_imports = true
 
 [tool.pydantic-mypy]
--- a/src/api/__init__.py
+++ b/src/api/__init__.py
@@ -1,0 +1,6 @@
+"""HTTP API surface for the ingest platform.
+
+Currently exposes the admin-only pipeline-reset endpoints (issue #70) that
+wrap the reset CLI surface (issue #9) so the isnad-graph admin reset UI can
+drive pipeline resets over HTTP. The app factory lives in :mod:`src.api.app`.
+"""
--- a/src/api/app.py
+++ b/src/api/app.py
@@ -1,0 +1,44 @@
+"""FastAPI app factory for the ingest-platform admin HTTP surface.
+
+Exposes the admin-only reset endpoints (issue #70). The reset router is
+mounted under ``/admin`` with ``Depends(require_admin)`` applied at the
+router level, so every reset route is admin-only — there is no
+unauthenticated path to a destructive operation.
+
+Run locally with::
+
+    uvicorn src.api.app:create_app --factory --port 8002
+"""
+
+from __future__ import annotations
+
+from fastapi import Depends, FastAPI
+
+from src.api.auth import require_admin
+from src.api.reset_router import router as reset_router
+
+
+def create_app() -> FastAPI:
+    """Construct the FastAPI application."""
+    app = FastAPI(
+        title="isnad-ingest admin API",
+        description="Admin-only pipeline operations (reset surface).",
+        version="0.1.0",
+    )
+
+    @app.get("/healthz", tags=["health"])
+    def healthz() -> dict[str, str]:
+        """Unauthenticated liveness probe."""
+        return {"status": "ok"}
+
+    # Admin guard is applied once at the mount point — covers every reset route.
+    app.include_router(
+        reset_router,
+        prefix="/admin",
+        dependencies=[Depends(require_admin)],
+    )
+
+    return app
+
+
+app = create_app()
--- a/src/api/auth.py
+++ b/src/api/auth.py
@@ -1,0 +1,176 @@
+"""Admin authentication for the reset HTTP surface.
+
+Mirrors the isnad-graph admin restriction so the *same* user-service RS256
+JWT (validated against user-service's JWKS) authorizes a reset here as on the
+isnad-graph admin dashboard. The reset endpoints are destructive, so the
+guard is admin-only — a valid non-admin token is rejected with 403.
+
+This is a deliberately slimmed copy of isnad-graph's
+``src/api/middleware.py`` + ``src/api/auth.py`` auth path: JWKS fetch with a
+TTL cache, RS256 verification with a single key-rotation retry, role
+resolution from the ``roles`` claim, and a ``require_admin`` FastAPI
+dependency. Kept self-contained (no shared package) because the two services
+are independent repos.
+"""
+
+from __future__ import annotations
+
+import logging
+import time
+from enum import StrEnum
+
+import httpx
+from fastapi import HTTPException, Request
+from jose import JWTError, jwt
+
+from src.config import get_settings
+
+logger = logging.getLogger(__name__)
+
+
+class Role(StrEnum):
+    """User roles with hierarchical privilege levels (matches isnad-graph)."""
+
+    VIEWER = "viewer"
+    EDITOR = "editor"
+    MODERATOR = "moderator"
+    ADMIN = "admin"
+
+
+ROLE_HIERARCHY: dict[Role, int] = {
+    Role.VIEWER: 0,
+    Role.EDITOR: 1,
+    Role.MODERATOR: 2,
+    Role.ADMIN: 3,
+}
+
+# user-service role string -> local Role. Matches isnad-graph's map so role
+# resolution is identical across services.
+_USER_SERVICE_ROLE_MAP: dict[str, Role] = {
+    "admin": Role.ADMIN,
+    "moderator": Role.MODERATOR,
+    "researcher": Role.EDITOR,
+    "editor": Role.EDITOR,
+    "reader": Role.VIEWER,
+    "viewer": Role.VIEWER,
+    "trial": Role.VIEWER,
+}
+
+
+def _resolve_role(roles: list[str]) -> Role:
+    """Pick the highest-privilege role from the JWT ``roles`` claim."""
+    best = Role.VIEWER
+    best_level = 0
+    for r in roles:
+        mapped = _USER_SERVICE_ROLE_MAP.get(r.lower(), Role.VIEWER)
+        level = ROLE_HIERARCHY.get(mapped, 0)
+        if level > best_level:
+            best = mapped
+            best_level = level
+    return best
+
+
+# ---------------------------------------------------------------------------
+# JWKS-based RS256 JWT validation (mirrors isnad-graph/src/api/auth.py)
+# ---------------------------------------------------------------------------
+
+_jwks_cache: dict[str, object] | None = None
+_jwks_fetched_at: float = 0.0
+
+
+def _get_jwks_url() -> str:
+    base = get_settings().auth.user_service_url.rstrip("/")
+    return f"{base}/.well-known/jwks.json"
+
+
+def fetch_jwks() -> dict[str, object]:
+    """Fetch JWKS from user-service, using a TTL-based in-memory cache.
+
+    Raises ``httpx.HTTPError`` if the JWKS endpoint is unreachable.
+    """
+    global _jwks_cache, _jwks_fetched_at  # noqa: PLW0603
+    ttl = get_settings().auth.user_service_jwks_cache_ttl
+    now = time.monotonic()
+    if _jwks_cache is not None and (now - _jwks_fetched_at) < ttl:
+        return _jwks_cache
+
+    resp = httpx.get(_get_jwks_url(), timeout=10.0)
+    resp.raise_for_status()
+    _jwks_cache = resp.json()
+    _jwks_fetched_at = now
+    return _jwks_cache
+
+
+def invalidate_jwks_cache() -> None:
+    """Force the next ``fetch_jwks`` call to re-fetch from user-service."""
+    global _jwks_cache, _jwks_fetched_at  # noqa: PLW0603
+    _jwks_cache = None
+    _jwks_fetched_at = 0.0
+
+
+def verify_user_service_token(token: str) -> dict[str, object]:
+    """Validate an RS256 JWT issued by user-service using its JWKS.
+
+    Invalidates the cache and retries once on signature failure to handle
+    key rotation. Returns the decoded payload.
+
+    Raises:
+        ValueError: token invalid, expired, or bad signature.
+        httpx.HTTPError: JWKS endpoint unreachable.
+    """
+    jwks = fetch_jwks()
+    try:
+        payload: dict[str, object] = jwt.decode(token, jwks, algorithms=["RS256"])
+    except JWTError:
+        logger.info("JWKS verification failed, invalidating cache and retrying")
+        invalidate_jwks_cache()
+        jwks = fetch_jwks()
+        try:
+            payload = jwt.decode(token, jwks, algorithms=["RS256"])
+        except JWTError as exc:
+            raise ValueError(f"Invalid token: {exc}") from exc
+    return payload
+
+
+# ---------------------------------------------------------------------------
+# FastAPI dependency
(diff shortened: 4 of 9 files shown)


parametrization:
**Reviewer verdict — secondary review**

Requestor: Sayed Reza
Requestee: Léopold Mbongo
RequestOrReplied: Approved
TechDebt: 3 non-blocking hardening items (see below): (1) no `aud`/`iss` claim verification in `verify_user_service_token` (mirrors isnad-graph by design); (2) `source` guard rejects `/` but allows `..` (not exploitable in object-store key space, but a positive allowlist would be tidier); (3) `python-jose[cryptography]>=3.3` floor is low (3.3.0 had CVE-2024-33663/33664) — uv.lock pins the fixed 3.5.0, so resolved tree is clean; raise floor to >=3.4 for belt-and-suspenders.

---

Reviewed independently at head 966b486 (base deployments/phase-4/wave-3). My angle: destructive-safety + auth, since this surface can wipe MinIO/B2/Kafka/Neo4j/PG.

**1. Admin enforcement is real server-side, on every route — PASS.** The guard is wired at the router mount in `src/api/app.py`: `app.include_router(reset_router, prefix="/admin", dependencies=[Depends(require_admin)])`. A router-level dependency runs on **every** route (stage/source/full); there is no per-route opt-out and no unauthenticated path to a destructive op (only `GET /healthz` is open, and it lives outside the router). `require_admin` traces to a genuine JWKS validation path, not a test mock: `verify_user_service_token` calls `jwt.decode(token, jwks, algorithms=["RS256"])` against a JWKS fetched from `user-service/.well-known/jwks.json` (TTL-cached, single key-rotation retry). Algorithm is pinned to RS256 only — closes alg=none and RS→HS key-confusion. It enforces `type == "access"`, `sub` is a str, and `_resolve_role(roles) == Role.ADMIN`. The test suite overrides only the `get_resetter`/`get_data_dir` deps (infra isolation); it does **not** override the auth dependency, so production wiring is the real guard. Status mapping is correct: 401 missing/invalid/expired token + wrong type, 403 valid-but-non-admin, 503 JWKS unreachable (`httpx.HTTPError`). Expiry is enforced (python-jose verifies `exp` by default → JWTError → 401).

**2. OBLITERATE gate + dry-run safety — PASS.** `reset_full` rejects with 400 when `confirmation != "OBLITERATE"` **before** any call to `_run_reset`, so no wipe can begin without the token. dry_run is **structurally** side-effect-free, not merely by convention: the dry-run branch calls `write_dry_run_audit(scope, *, confirmation_method, data_dir)`, whose signature receives **no** infra adapters — only the real `resetter.reset()` path holds object_store/kafka/neo4j/pg. A dry-run therefore *cannot* touch infra; it only writes an audit entry (matching the CLI). Strong guarantee.

**3. Input validation — PASS.** `stage` is enum-constrained: `ResetScope.stage_scope` raises `ValueError` (→422) when `stage not in STAGE_PREFIXES`. `source` carries a path-traversal/scope-escape guard: `source_scope` raises (→422) on empty-or-contains-`/`, preventing both a full-stage wipe via empty source and an injected extra path segment. The Pydantic request models are `frozen` + `extra="forbid"`, so unknown fields are rejected (→422). Error codes are correct throughout (422 scope-validation, 400 missing OBLITERATE token, 401/403/503 auth). The residual `..`-without-slash case is non-exploitable: object-store keys are literal, not filesystem paths — `raw/../` matches no keys and B2/S3 do not normalize `..` — captured as TechDebt (2) above.

**4. No credential/internal leakage — PASS.** `ResetResponse` exposes only `level`, `dry_run`, `confirmation_method`, `audit_entry_path`, and the resetter's `to_summary()` — no connection strings, tokens, or adapter internals. Auth failures return generic details. Deps reasonable: `fastapi>=0.115`, `uvicorn[standard]>=0.34`, `python-jose[cryptography]>=3.3` (the `[cryptography]` extra selects the pyca backend for RS256). uv.lock pins exact versions (python-jose 3.5.0 — current, unflagged).

**5. CI at head — security-audit RED is known not-this-PR noise.** lint-and-typecheck, precommit-ci-sync, and test are all GREEN. `security-audit` fails on a single finding: `pip 26.1.1 / PYSEC-2026-196` (fix 26.1.2) — the runner's bundled pip, a transitive/build tool. This PR's uv.lock adds **no** `pip` pin (the 279 lock lines are fastapi/uvicorn/python-jose + their transitive deps), and none of the introduced runtime deps are flagged. This is the recurring pip-audit `--strict` advisory-DB drift, not a regression here. Recommend a **separate** one-line uv.lock pip bump to 26.1.2 to clear the org-wide noise; do not fold it into this scoped PR.

Net: the destructive surface is correctly gated — real admin-only enforcement on every route, OBLITERATE-before-wipe, structurally inert dry-runs, and constrained inputs. Approved.

parametrization:
**Reviewer verdict — primary review**

Requestor: Petra Vidović
Requestee: Léopold Mbongo
TechDebt: (1) confirmation_method audit channel reuses the TTY-typed "interactive" value for HTTP-triggered obliterates — recommend a distinct SIEM value ("api") as a follow-up that touches the merged reset.py contract; (2) no test covers the 503 JWKS-unreachable auth path.

---

Reviewed at head `966b486` against the real diff (not the PR body). One blocking item; auth and the OBLITERATE gate are genuinely solid.

### What holds up (verified server-side, not via the PR framing)

- **Auth is genuinely enforced, not a mock bypass.** `require_admin` is mounted once at the router (`include_router(reset_router, prefix="/admin", dependencies=[Depends(require_admin)])`), so every reset route is guarded with no unauthenticated path to a destructive op. The four auth tests use `_client_with_real_guard`, which does NOT override `require_admin` — the real dependency runs; only the JWKS *network fetch* (`verify_user_service_token`) is monkeypatched. So the missing-header → 401, non-access-type → 401, viewer → 403, admin → 200 assertions exercise the actual guard logic. The 503 (JWKS unreachable) and 401 (bad token type / non-str sub) branches are all present in `auth.py`. This is the right shape and avoids the test-mock-masks-the-guard trap.
- **OBLITERATE is checked before any destructive work.** `reset_full` rejects `confirmation != "OBLITERATE"` with 400 *before* calling `_run_reset`; a missing field is 422 (pydantic required + `extra="forbid"`). Both reject-paths assert the resetter was never called. Correct ordering.
- **Scope dispatch + audit per call are correct.** stage→`ResetScope.stage_scope`, source→`source_scope`, full→`full_scope`; unknown stage / bad source → 422 before any reset. Real path goes through `PipelineResetter.reset`; dry-run through `write_dry_run_audit`; both emit an audit entry. The `confirmation_method` values used ("not-required" for stage/source, "interactive" for full) are both members of the `CONFIRMATION_METHODS` frozenset, so `_validate_confirmation_method` will not raise.
- Deps are pinned to sane lower bounds (fastapi>=0.115, uvicorn[standard]>=0.34, python-jose[cryptography]>=3.3, httpx>=0.27) with uv.lock updated.

### Blocking: dry-run opens live infra connections (contract says it must not)

The dry-run contract — "validates scope + writes a dry-run audit entry, touches no infra" — does not hold at runtime, and the unit suite masks it.

`get_resetter` is declared as an unconditional FastAPI dependency on all three reset routes (`resetter: Annotated[PipelineResetter, Depends(get_resetter)]`). FastAPI resolves every declared dependency *before* invoking the handler, regardless of which branch the handler takes. So even for `dry_run: true`, `get_resetter` → `cli._build_reset_clients(settings)` runs, and that function:

- reads `os.environ["KAFKA_BOOTSTRAP_SERVERS"]` → `KeyError` (→ 500) if the env var is unset;
- constructs `KafkaAdminClient(bootstrap_servers=bootstrap)` → kafka-python bootstraps the cluster connection on construction;
- calls `psycopg.connect(pg_cfg.dsn)` → opens a live Postgres connection immediately.

So a dry-run preview — the safe affordance ig#970's admin UI will call first — actually opens Postgres + Kafka connections and hard-requires Kafka config. `_run_reset` never *uses* `resetter` on the dry-run branch, but the connection is already open by the time we get there.

The tests don't catch this because the `client` fixture overrides `get_resetter` with `_FakeResetter` (and `_client_with_real_guard` does too). `test_dry_run_writes_audit_without_invoking_resetter` only asserts `fake_resetter.calls == []` — i.e. that `.reset()` wasn't *called* — not that no client was *built*. This is exactly the test-mock-injection-masks-production-failure pattern.

**Required fix:** make resetter construction lazy so the dry-run path never triggers live-client construction. The cleanest shape is to inject a factory rather than an instance — e.g. a `Depends(get_resetter_factory)` returning `Callable[[], PipelineResetter]`, and call it only in the non-dry-run branch of `_run_reset`. `get_data_dir` stays a normal dependency (it's infra-free and is what the dry-run audit write needs). A regression test should assert the factory is *not* invoked when `dry_run=true` (e.g. a sentinel factory that raises if called).

### Your flagged question — confirmation_method = "interactive" for the API path

My call: **acceptable for this PR, do not block on it — but introduce a distinct value as a tracked follow-up.**

Reasoning: "interactive" is a contract-valid member of `CONFIRMATION_METHODS`, so nothing crashes and the audit entry is still written truthfully as far as the schema is concerned. But the entire purpose of that field is SIEM/forensic fidelity — *how* the obliterate was authorized — and an HTTP request carrying an OBLITERATE token is a materially different security event from a human typing OBLITERATE at a TTY. Collapsing both into "interactive" loses the channel distinction precisely where it matters most (the full-obliterate path). A distinct "api" value would be a one-line additive change to the frozenset (the validator already reads it generically).

I'm making it a follow-up rather than a blocker because adding the value touches the already-merged reset.py contract (#9) and the isnad-graph SIEM/dashboard side should be updated in lockstep — that deserves its own small issue, not a quiet widening folded into this PR. Logged as TechDebt above.

### Minor (non-blocking)

- No test exercises the 503 path (JWKS unreachable). The branch exists in `require_admin` (`except httpx.HTTPError: → 503`); a stub raising `httpx.HTTPError` would close the contract item the brief lists.

Happy to re-review promptly once the dry-run wiring is made lazy. The auth and OBLITERATE work is good — the block is narrowly the eager client construction on the dry-run path.

parametrization:
**Reviewer verdict — primary re-review**

Requestor: Petra Vidović
Requestee: Léopold Mbongo
RequestOrReplied: Approved
TechDebt: Two tracked follow-ups, neither blocking — (1) introduce a distinct "api" value in CONFIRMATION_METHODS so HTTP-triggered obliterates are SIEM-distinguishable from a TTY-typed confirmation (touches the merged reset.py #9 contract + isnad-graph SIEM side, own issue); (2) security-audit pip advisory-DB drift on the branch (not introduced here).

---

Re-reviewed at head `b361a17` against the real diff (`git diff 966b486 b361a17` — scope is exactly deps.py + reset_router.py + tests, so the rest of my prior review stands).

**Blocking item is genuinely fixed.** `get_resetter` (an unconditional instance dependency) is replaced by `get_resetter_factory`, which returns an *unevaluated* `_build` closure. The live `_build_reset_clients` call — the one that reads `KAFKA_BOOTSTRAP_SERVERS` and eagerly opens Kafka + Postgres connections — now lives inside `_build` and runs only when the factory is *invoked*. `_run_reset` invokes it solely in the `else` (non-dry-run) branch; the dry-run branch goes straight to `write_dry_run_audit` and never calls it. Traced all three routes (stage/source/full): each declares `Depends(get_resetter_factory)`, and the `/full` OBLITERATE 400-check still precedes `_run_reset`, so a rejected or dry-run call constructs no clients.

**The regression test proves non-construction, not just non-call.** `_FactorySpy.__call__` increments `build_count` — i.e. it counts how many times a route asked to *build* the live resetter, which is the precise property I asked for (the prior suite only asserted `.reset()` wasn't called, which masked the eager construction). `test_dry_run_never_builds_the_resetter`, parametrized across stage/source/full, asserts `build_count == 0` on every dry-run; `test_real_reset_builds_the_resetter_once` asserts `build_count == 1` on a real reset. The spy is wired in via `dependency_overrides[get_resetter_factory]`, so it stands in for `_build` faithfully — a `build_count == 0` genuinely proves `_build_reset_clients` is never reached on a dry-run.

**503 gap closed.** `test_jwks_unreachable_returns_503` raises `httpx.ConnectError` (an `httpx.HTTPError` subclass) from `verify_user_service_token` and asserts 503 with the resetter untouched — exercises the real `require_admin` branch.

Everything I verified solid in the first pass is unchanged: auth is enforced server-side at the router mount (real guard run in the auth tests, only the JWKS fetch stubbed), the OBLITERATE gate rejects before any work (400 wrong / 422 missing), and scope dispatch + per-call audit are correct. Reported local run: tests/test_api/ 17 passed (was 12), full suite 622 passed / 3 skipped, ruff + mypy clean.

On the `confirmation_method` question my position is unchanged and already captured as TechDebt above: reusing "interactive" is contract-valid and fine to ship; the distinct "api" value is a follow-up, not a blocker.

Approved. Nice, tightly-scoped fix.

PR #67, opened 2026-06-11
test(integration): full dedup→enrich→normalize→ingest worker-chain E2E (Refs noorinalabs-main#136)

parametrization:
## Summary
Refs noorinalabs-main#136 — full worker-chain integration scenarios for `dedup → enrich → normalize → ingest`, delivered as **two legs** per the team-lead spec:

1. **CI-durable in-process chain** (`tests/workers/test_worker_chain_inprocess.py`) — drives all four stages in-process on the faithful fake stack (`FakeS3Client` → `ObjectStore` + parquet builders), with a **faithful** in-process Neo4j fake (`FakeNeo4j`) that tracks MERGEd node ids so the ingest edge MERGE's `OPTIONAL MATCH` endpoint resolution is reproduced. Runs in the default `pytest -m "not integration"` lane — **CI proves the worker-chain contract with zero Docker**.
2. **Docker-gated graph-load tail** (`tests/integration/test_worker_chain_e2e.py`) — the same four stages through **real Kafka + MinIO + Neo4j** testcontainers, terminating in a real Neo4j MERGE. `@pytest.mark.integration`, runs via `make test-integration`; CI-runner-flake leg tracked at noorinalabs-main#631.

`Closes` does not fire cross-repo → **Refs noorinalabs-main#136**.

## What this fills vs the existing tests
- **#55 `test_kafka_worker_e2e.py`** seeds at `pipeline.dedup.done` — the dedup worker never runs. Both legs here run the dedup hop.
- **ig#62 / main#139 `test_e2e_inprocess.py`** asserts node **counts**, not ids.
- **Pre-existing `test_processor_chain_propagates_batch_id`** stops at normalize and asserts only batch-id propagation. The in-process leg here carries the chain through the **ingest graph-load tail** and asserts the real graph end-state.

## Contracts pinned
- **Hadith-id parity (#63)** — corpus-prefixed fixture (`sunnah:bukhari:1:1`) MERGEs as `hdt:sunnah:bukhari:1:1` (the batch loader's `hdt:{source_id}`), never the doubled `hdt:sunnah:sunnah:...`. Guarded in BOTH legs. (Stacked on #66, the #63 fix.)
- **normalize→ingest manifest contract** — full-label fan-out consumed without tripping the fail-loud edge guard.
- **Chain idempotency on replay** — deterministic ids → identical node set.
[...]

Issue #74, opened 2026-06-12
refactor(reset): distinct confirmation_method="api" for HTTP-driven full reset (SIEM clarity)


parametrization:
## Context

Follow-up from ingest-platform#70 / PR #73 (admin HTTP reset endpoints). The `POST /admin/reset/full` endpoint requires the `OBLITERATE` confirmation token in the request body, and records the reset's `confirmation_method` for SIEM attribution.

To avoid modifying the merged reset.py #9 contract in a scoped feature PR, #73 **reused the existing `"interactive"` value** for the API-supplied OBLITERATE token (rationale: a human typed `OBLITERATE` into the admin form). Petra flagged this as contract-valid but not ideal for SIEM: `"interactive"` semantically means a TTY `input()` confirm, whereas this came over HTTP.

## Proposal

Add a distinct `confirmation_method = "api"` (or `"api-token"`) value so SIEM can cleanly distinguish:
- `interactive` — operator typed OBLITERATE at the CLI prompt (TTY)
- `yes-flag` — CLI `--yes` runbook automation
- `not-required` — non-obliterate scope
- **`api`** — OBLITERATE token supplied via the admin HTTP endpoint

## Why this is its own issue (not folded into #73)

This touches **two merged contracts in lockstep**:
1. `src/pipeline/reset.py` `CONFIRMATION_METHODS` frozenset (merged via #9) — add the new value + its reset.py tests.
2. The isnad-graph **SIEM side** — any dashboards/queries that filter on `confirmation_method` must learn the new value at the same time, or the admin-driven full resets become invisible/misclassified in SIEM.

## Acceptance

- `"api"` added to `CONFIRMATION_METHODS`; `/admin/reset/full` records it instead of `"interactive"`.
- reset.py + reset-endpoint tests updated.
- isnad-graph SIEM consumers updated/confirmed to recognize the new value (coordinate, don't drift).

Provenance: ingest-platform#70, PR #73 (merged 803dd6d), Petra Vidović review.

parametrization:
Merged to wave branch deployments/phase-4/wave-4 via PR #78. confirmation_method=api for HTTP reset (SIEM).

PR #79, opened 2026-06-12
fix(reset): _delete_prefix works against real MinIO/B2 (MissingContentMD5) (#69)

### LINKED ISSUE #69: bug(pipeline/reset): _delete_prefix bulk delete_objects → MissingContentMD5 on real MinIO/B2 (botocore≥1.36); reset broken vs real object store
## Symptom

`isnad-ingest` admin stage-reset fails against any real S3-compatible object store (MinIO, Backblaze B2) with:

```
MissingContentMD5
```

Surfaced 2026-06-11 by the #136 worker-chain E2E (PR #67) — its real-MinIO reset→rerun scenario hit this and is currently `xfail(strict=True)` pending this fix.

## Root cause

`src/pipeline/reset._delete_prefix` (reset.py:~408) calls the **bulk** `client.delete_objects(...)` (S3 `DeleteObjects`). Under **botocore ≥ 1.36**, boto3 no longer auto-adds the `Content-MD5` / `x-amz-checksum` header that S3-compatible stores require for the bulk `DeleteObjects` API, so MinIO/B2 reject the request. AWS S3 itself is more lenient, which is why it wasn't caught against AWS.

The unit suite missed it because `FakeS3Client.delete_objects` is a **no-op** — a test-mock-masks-production-failure (same class as the other two bugs this wave: da#77 null-MERGE, ig#63 id double-prefix).

## Fix

Use the per-object `ObjectStore.delete_object` (object_store.py:~152) — which the codebase already uses elsewhere — to delete each key under the prefix, instead of the bulk `delete_objects`. (Alternatively, set the required `ChecksumAlgorithm`/`Content-MD5` on the bulk call — but per-object delete matches existing patterns and is simpler.)

## Acceptance

- [ ] `reset._delete_prefix` deletes prefix contents against a real MinIO/B2 (no `MissingContentMD5`)
- [ ] PR #67's real-MinIO reset→rerun scenario flips from `xfail` to pass (remove the strict-xfail once green)
- [ ] a test exercises the delete path against a real (testcontainers) store, not the no-op fake

## Notes
Surfaced (not introduced) during P4W2 #136. Reset is broken vs ANY real S3/MinIO/B2 store today — worth prompt scheduling.

parametrization:
Closes #69

## Problem
`src/pipeline/reset._delete_prefix` (reset.py:~390) wiped a stage's objects during admin reset via the **bulk** `client.delete_objects(...)` (S3 `DeleteObjects`). Under **botocore ≥ 1.36**, boto3 no longer auto-adds the `Content-MD5`/checksum header that the bulk op requires, so real S3-compatible stores (**MinIO, Backblaze B2**) reject the request with **`MissingContentMD5`**. AWS S3 is lenient, which is why it wasn't caught against AWS. Net: admin stage-reset was broken against any real object store — it only "passed" because the unit suite's `FakeS3Client` is an in-memory no-op (a test-mock-masks-production-failure, same class as da#77 null-MERGE and ig#63 id double-prefix this wave).

## Fix
Delete each key under the prefix with the per-object **`delete_object`** (`DeleteObject`) — which carries no `Content-MD5` requirement and matches the per-object delete `ObjectStore.rename_object` already uses (`workers/lib/object_store.py:152`). The paginated `list_objects_v2` continuation loop is unchanged, so large/truncated listings still page through fully.

```diff
-        if contents:
-            to_delete = [{"Key": obj["Key"]} for obj in contents]
-            client.delete_objects(Bucket=store.bucket, Delete={"Objects": to_delete, "Quiet": True})
-            deleted += len(to_delete)
+        for obj in contents:
+            client.delete_object(Bucket=store.bucket, Key=obj["Key"])
+            deleted += 1
```

## How it's tested (against a real-ish client, not the masking mock)
- **New `tests/integration/test_minio_reset_delete.py`** — drives `_delete_prefix` against a **real testcontainers MinIO** container (single container, fast; vs the 3-container chain E2E). Case 1: seeds two keys under `normalized/` + a sibling `raw/` key, asserts the prefix is emptied, the sibling survives, count is correct, **no `MissingContentMD5`**. Case 2: seeds >1000 keys to exercise the continuation-token pagination path against the real store.
  - **Bug-catch verified:** temporarily reverting `_delete_prefix` to the old bulk `delete_objects` makes this test fail with exactly `botocore.exceptions.ClientError: ... (MissingContentMD5) ... DeleteObjects ... Missing required header ... Content-Md5`. With the per-object fix it passes. So the test genuinely guards the regression — it is not a no-op like the old fake.
- **Un-xfail'd** the #67 real-MinIO reset→rerun scenario in `tests/integration/test_worker_chain_e2e.py` (`test_admin_stage_reset_wipes_only_normalized_prefix_on_real_minio`) — the full-chain regression guard the issue referenced. Removed the strict-xfail and refreshed its docstring.
- **Updated the unit `FakeS3Client`** (`tests/test_pipeline/test_reset.py`) to model `delete_object` rather than bulk `delete_objects`, so the fake tracks the production surface and can't re-mask the bug.

### Test results (this branch, run locally with Docker)
- `tests/test_pipeline/test_reset.py` — **26 passed**
- `tests/integration/test_minio_reset_delete.py` (real MinIO) — **2 passed**
- ruff 0.15.11 format `--check` + lint — clean; `mypy src/pipeline/reset.py` — clean.

## Serial-merge note (for the wave coordinator)
This PR touches `src/pipeline/reset.py` **only inside `_delete_prefix` (~L390–417)** and `_FakeS3Client` in the unit test. ingest#74 (Kalinda) touches the `CONFIRMATION_METHODS` constant / `_validate_confirmation_method` (~L380, a different region of the same file) — **non-overlapping**, so the two should merge cleanly in either order. Flagging for sequencing per the brief.

## Reviewers
@Léopold Mbongo (primary) · @Petra Vidović (secondary)

## Test Plan
- [x] Per-object delete path emties a prefix against real MinIO with no `MissingContentMD5`
- [x] Sibling prefixes survive a scoped delete; deleted-count is exact
- [x] Pagination (>1000 keys) deletes every object
- [x] #67 real-MinIO reset→rerun scenario flips xfail→pass
- [x] Unit suite green with the `delete_object` fake
- [ ] (CI) integration lane runs the new test where Docker is available (`make test-integration`)

### DIFF
--- a/src/pipeline/reset.py
+++ b/src/pipeline/reset.py
@@ -391,7 +391,15 @@ def _delete_prefix(store: _ObjectStoreProto, prefix: S3Prefix) -> int:
     """Delete every object under ``prefix`` in ``store.bucket``.
 
     Returns the number of objects deleted. Uses the paginated ``list_objects_v2``
-    surface so very large prefixes don't blow past a single-response limit.
+    surface so very large prefixes don't blow past a single-response limit, and
+    removes each key with the per-object ``delete_object`` (S3 ``DeleteObject``)
+    rather than the bulk ``delete_objects`` (``DeleteObjects``). The bulk op
+    requires a ``Content-MD5``/checksum header that boto3 (botocore>=1.36) no
+    longer adds by default, so real S3-compatible stores (MinIO, Backblaze B2)
+    reject it with ``MissingContentMD5`` — AWS S3 happens to be lenient, which
+    is why it slipped through. ``DeleteObject`` carries no such requirement and
+    matches the per-object delete ``ObjectStore.rename_object`` already uses
+    (#69).
     """
     client = store.client
     deleted = 0
@@ -403,10 +411,9 @@ def _delete_prefix(store: _ObjectStoreProto, prefix: S3Prefix) -> int:
             kwargs["ContinuationToken"] = continuation
         response = client.list_objects_v2(**kwargs)
         contents = response.get("Contents", []) or []
-        if contents:
-            to_delete = [{"Key": obj["Key"]} for obj in contents]
-            client.delete_objects(Bucket=store.bucket, Delete={"Objects": to_delete, "Quiet": True})
-            deleted += len(to_delete)
+        for obj in contents:
+            client.delete_object(Bucket=store.bucket, Key=obj["Key"])
+            deleted += 1
         if not response.get("IsTruncated"):
             break
         continuation = response.get("NextContinuationToken")
--- a/tests/integration/test_minio_reset_delete.py
+++ b/tests/integration/test_minio_reset_delete.py
@@ -1,0 +1,121 @@
+"""Focused real-MinIO regression guard for ``reset._delete_prefix`` (#69).
+
+``src/pipeline/reset._delete_prefix`` wipes a stage's objects during an admin
+reset. It used the bulk ``delete_objects`` (S3 ``DeleteObjects``) op, which
+boto3 (botocore>=1.36) no longer auto-signs with a ``Content-MD5``/checksum
+header — so real S3-compatible stores (MinIO, Backblaze B2) reject it with
+``MissingContentMD5``. AWS S3 is lenient, and the unit suite's ``FakeS3Client``
+is in-memory, so the defect only shows against a *real* store.
+
+The full ``dedup→…→ingest`` chain E2E
+(``test_worker_chain_e2e.test_admin_stage_reset_wipes_only_normalized_prefix_on_real_minio``)
+also guards this, but stands up THREE containers (Kafka, MinIO, Neo4j). This
+module isolates the object-store delete path against a SINGLE MinIO container so
+the regression is caught fast, without the chain or the Kafka/Neo4j stack — a
+test that would have failed (``MissingContentMD5``) before the per-object
+``delete_object`` fix and passes after.
+
+Runtime requirement
+-------------------
+``@pytest.mark.integration`` — needs Docker (testcontainers spins up a MinIO
+container). Excluded from the default ``pytest -m "not integration"`` run;
+runs via ``make test-integration``. ``importorskip`` on
+``testcontainers.minio`` / ``boto3`` keeps a Docker-less lane collecting cleanly.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Iterator
+
+import pytest
+
+pytest.importorskip("testcontainers.minio")
+pytest.importorskip("boto3")
+
+import boto3  # noqa: E402
+from botocore.config import Config  # noqa: E402
+from testcontainers.minio import MinioContainer  # noqa: E402
+
+from src.pipeline.reset import _delete_prefix  # noqa: E402
+from workers.lib.object_store import ObjectStore  # noqa: E402
+
+pytestmark = pytest.mark.integration
+
+TEST_BUCKET = "noorinalabs-pipeline"
+
+
+@pytest.fixture(scope="module")
+def minio_container() -> Iterator[MinioContainer]:
+    with MinioContainer(access_key="minio", secret_key="minio12345") as minio:
+        yield minio
+
+
+@pytest.fixture
+def minio_store(minio_container: MinioContainer) -> ObjectStore:
+    """Real ``ObjectStore`` (boto3) on the MinIO container, bucket pre-created.
+
+    Path-style addressing because MinIO does not serve virtual-host bucket
+    subdomains.
+    """
+    cfg = minio_container.get_config()
+    client = boto3.client(
+        "s3",
+        endpoint_url=f"http://{cfg['endpoint']}",
+        aws_access_key_id=cfg["access_key"],
+        aws_secret_access_key=cfg["secret_key"],
+        region_name="us-east-1",
+        config=Config(s3={"addressing_style": "path"}),
+    )
+    try:
+        client.create_bucket(Bucket=TEST_BUCKET)
+    except client.exceptions.BucketAlreadyOwnedByYou:
+        pass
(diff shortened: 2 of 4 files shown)


parametrization:
Requestor: Léopold Mbongo
Requestee: Fatima Bensalah
RequestOrReplied: Approved

Reviewed at PR HEAD via `gh pr diff` + files API. Approving — the fix is correct, the guard is real, and CI is green.

## What I verified

**(a) Bulk → per-object delete.** `_delete_prefix` (reset.py L411-416) now loops `client.delete_object(Bucket=..., Key=obj["Key"])` per key instead of one bulk `client.delete_objects(Delete={...})`. `DeleteObject` carries no `Content-MD5`/checksum requirement, so the `MissingContentMD5` rejection that MinIO/B2 raise under botocore>=1.36 disappears. This matches the per-object surface `ObjectStore.rename_object` already uses. Correct root-cause fix, not a header workaround.

**(b) New real-MinIO test genuinely fails on the old path — not a masked no-op.** `tests/integration/test_minio_reset_delete.py` drives `_delete_prefix` against a real testcontainers MinIO via a real boto3 client (path-style addressing for MinIO). It seeds objects, calls the production `_delete_prefix`, and asserts `deleted == 2` with no exception. Against the old bulk `delete_objects` this raises `MissingContentMD5` before any assert, so it genuinely flips fail→pass. Unlike the in-memory `FakeS3Client`, there is no mock to swallow the error. PR body documents the manual revert-and-fail check producing exactly `(MissingContentMD5) ... DeleteObjects ... Missing required header ... Content-Md5`.

**(c) Unit fake updated so it can't re-mask.** `_FakeS3Client` (tests/test_pipeline/test_reset.py) drops `delete_objects` entirely and models `delete_object(*, Bucket, Key)`. The fake now tracks the exact production surface — if someone reverts to the bulk call, the unit suite breaks too (AttributeError), not just the integration lane. This closes the test-mock-masks-production-failure class for this path.

**(d) Xfail removed.** The `@pytest.mark.xfail(strict=True)` on `test_admin_stage_reset_wipes_only_normalized_prefix_on_real_minio` (the real-MinIO reset→rerun chain scenario) is removed and its docstring refreshed to record the former-xfail history. `strict=True` means it would now XPASS-fail if the bug regressed, so removing it is the right move once the fix is in.

**(e) Pagination preserved.** The `list_objects_v2` continuation-token loop (`ContinuationToken` / `IsTruncated` / `NextContinuationToken`) is unchanged. The new `test_delete_prefix_paginates_past_single_response_on_real_minio` seeds 1001 keys to exercise the >1000-key multi-page path against the real store and asserts all are deleted.

**CI green:** lint-and-typecheck, precommit-ci-sync, security-audit, test all pass. Integration lane (the two real-MinIO tests) is Docker-gated and excluded from the default `pytest -m "not integration"` CI run — the `importorskip` on testcontainers.minio/boto3 keeps the Docker-less lane collecting cleanly. That gating is expected and acceptable.

## Serial-merge / non-overlap
Confirmed non-overlapping with ingest#74. This PR touches `src/pipeline/reset.py` only inside `_delete_prefix` (~L391-417) plus the unit `_FakeS3Client`; #74 touches `CONFIRMATION_METHODS` / `_validate_confirmation_method` (~L380), a distinct region. The two merge cleanly in either order.

Nice work — the regression-guard discipline (real container, fail-on-old verification, fake updated to match) is exactly right.

parametrization:
Requestor: Petra Vidović
Requestee: Fatima Bensalah
RequestOrReplied: Approved

Secondary (tech-lead) review — verified via diff + CI + source read, did not run the local suite.

**Verdict: Approved.** Clean, correctly-scoped fix for a real defect.

Tech-lead lens:

- **(a) Per-object delete is correct, and the perf trade-off is right here.** `_delete_prefix` is a bounded admin reset op, not a hot path, so swapping the single bulk `DeleteObjects` for N `DeleteObject` calls per page is an acceptable cost. The alternative — keeping bulk delete and re-adding a `Content-MD5`/checksum header via a botocore config knob — would be more fragile (tied to botocore internals that already shifted under us in ≥1.36) and wouldn't match anything else in the codebase. Per-object is the simpler, more durable choice and, as I confirmed at `workers/lib/object_store.py:rename_object`, it matches the per-object `delete_object` the store already uses for copy+delete. Consistency win.

- **(b) Test layer is right.** The single-container testcontainers MinIO guard (`test_minio_reset_delete.py`) isolates the object-store delete path fast, and the strict-xfail removal on the 3-container chain E2E keeps the full-stack guard honest too. Critically the masking `_FakeS3Client` was also updated to model `delete_object` rather than the in-memory `delete_objects` no-op — so the fake can no longer re-hide this class of bug. That's the part that makes the fix stick (test-mock-masks-production-failure was the root cause). The pagination case (>1000 keys) against the real store is a good addition.

- **(c) Partial-delete error handling is sane — arguably improved.** If a mid-loop `delete_object` raises, the exception propagates and the count is partial, but the reset is idempotent (re-run resumes wiping the remaining keys), so that's the correct failure posture for an admin reset. Notably the old bulk path used `Quiet=True`, which silently swallowed per-object failures in the `DeleteObjects` response — so this change actually surfaces real delete errors instead of hiding them. No regression; a small improvement.

- **(d) Scope held.** Touches only `_delete_prefix` (~L390–417) and the unit fake. Non-overlapping with ingest#74's `CONFIRMATION_METHODS` region as flagged — serial-merge note is accurate.

CI: lint-and-typecheck, precommit-ci-sync, security-audit, test all green.

Nice work, Fatima — the bug-catch verification (reverting to bulk delete reproduces `MissingContentMD5`) is exactly the evidence I want to see on a mock-masking fix.

PR #85, opened 2026-06-13
fix(graph/load_edges): null-safe APPEARS_IN MERGE — SET in-book ordinal (#84)

### LINKED ISSUE #84: bug(graph/load_edges): APPEARS_IN MERGE embeds nullable hadith_number (load_edges.py:279) — null in-book ordinal aborts the write batch (ingest-platform sibling of da#77)
## Bug (verified at HEAD `src/graph/load_edges.py:279`, 2026-06-13)

`_APPEARS_IN_QUERY` embeds a **nullable** property inside the relationship MERGE pattern:

```cypher
MERGE (h)-[:APPEARS_IN {
    book_number: row.book_number,
    chapter_number: row.chapter_number,
    hadith_number: row.hadith_number      <-- nullable
}]->(c)
```

`hadith_number` is sourced from `_val(row, "hadith_number")` (line 336) and can be `None`. The pre-load validity filter (`_load_appears_in`, lines 321–323) only guards `source_id` / `collection_name` — it **never** guards `hadith_number`. Neo4j raises on `MERGE` with a null property in the pattern (`Cannot merge relationship using null property value`), which **aborts the entire write batch** — so a single scraped Hadith with no in-book ordinal kills the load for every Hadith in that batch. This is the same failure class as **da#73** (real scraped data, null in-book ordinal).

## Why this is not a dup of da#77

`da#77` fixed the **data-acquisition** batch loader's copy of this bug. This is the **ingest-platform worker-path loader** — a separate, still-unfixed sibling that runs in the deployed pipeline. The two loaders drifted (cf. da#65: the key should be the canonical `hadith_number_in_book`, not `hadith_number` — that drift is also present here and worth fixing in the same pass).

## Fix sketch

Move `hadith_number` out of the MERGE pattern and `SET` it after the MERGE (so a null no longer participates in relationship identity), or filter null-ordinal rows into the `skipped` path like `source_id`/`collection_name`. Mirror whatever da#77 landed for parity, and align the property key with the canonical `hadith_number_in_book` (da#65).

## Impact

Latent today only because the deployed pipeline has never run on staging (main#601 criterion #1) and toy `h-1` fixtures mask it — but it will abort the **first** real worker-driven load of scraped data. Blocks the ingest-platform staging-E2E gap filed alongside this.

## Provenance

Surfaced during the `main#601` P4W5 exit-drive verification. Owner disposition 2026-06-13 = Option A (file the gaps; remediation to Phase 5). No wave label — Phase-5 scoping candidate (could be pulled earlier as a discrete bug fix if a staging E2E run is scheduled sooner).

parametrization:
## Summary

Fixes the ingest-platform worker-path copy of the APPEARS_IN null-property loader bug (sibling of da#77).

`_APPEARS_IN_QUERY` embedded three nullable positional props inside the relationship MERGE pattern:

```cypher
MERGE (h)-[:APPEARS_IN {book_number: ..., chapter_number: ..., hadith_number: row.hadith_number}]->(c)
```

Neo4j refuses to MERGE a relationship with a null property value, so a single scraped hadith whose in-book ordinal is null raises `Cannot merge relationship using null property value` and **aborts the entire write batch** — every hadith in the batch fails to load. The `_load_appears_in` validity filter only guarded `source_id`/`collection_name`, never the ordinal, and `MockNeo4jClient` masked it in tests.

## Fix

- MERGE on the stable `(hadith, collection)` PAIR identity only; `SET` the positional props after the MERGE so a null never participates in relationship identity (null-safe).
- Use the canonical edge key `hadith_number_in_book` (mapped from `row.hadith_number`, the in-book ordinal — Book 1 Hadith 1), not the bare `hadith_number`. This aligns the batch loader with the streaming normalize path (`workers/normalize/processor.py`), `src/models/edges.py`, and `EDGE_PROPERTY_MAP`, which already use it (da#65). The batch loader was the lone holdout.
- Use the streaming path's `coalesce`-preserve contract so a re-run carrying a null ordinal never clobbers a value an earlier load set.

## Tests

- **`test_load_edges.py`** (mock suite): null-ordinal row reaches the batch (not skipped, `created==1`) + structural guards asserting the MERGE line carries no inline property map and the SET uses the canonical `coalesce` key. The mock cannot enforce the Neo4j MERGE-null rule, hence the structural guards.
- **`test_graph_loading.py`** (real Neo4j container): regression test loading a null-ordinal hadith batched with a non-null sibling; asserts both edges created (batch **not** aborted), the null edge has the key absent (Neo4j drops a null-SET prop), and the sibling carries the ordinal under the canonical key. **Verified locally against a live `neo4j:5` container** (`1 passed`).

Full non-integration suite: `645 passed`. ruff (pinned 0.15.11) + mypy clean.

## Test Plan

- [x] `ruff format --check` + `ruff check` (src/tests) — clean
- [x] `mypy src/` — clean
- [x] `pytest -m "not integration"` — 645 passed
- [x] `pytest -m integration` TestEdgeLoading (real Neo4j container) — 3 passed

Closes #84

Reviewers: Léopold Mbongo, Camila Restrepo

### DIFF
--- a/src/graph/load_edges.py
+++ b/src/graph/load_edges.py
@@ -273,15 +273,28 @@ def _load_narrated(
 # 3. APPEARS_IN — hadith -> collection
 # ---------------------------------------------------------------------------
 
+# Identity of an APPEARS_IN edge is the (hadith, collection) PAIR — a hadith
+# appears in a given collection at most once. The positional props (book /
+# chapter / in-book ordinal) describe WHERE in that collection and are not part
+# of its key. They are therefore SET after the MERGE, never inside the MERGE
+# pattern (ingest-platform sibling of da#77): Neo4j refuses to MERGE a
+# relationship with a null property value, and scraped hadiths leave the in-book
+# ordinal null until it is enriched, so a keyed-on-properties MERGE aborts the
+# whole write batch on real scraped data. SET is null-safe.
+#
+# The SET uses the streaming ingest path's coalesce-preserve contract
+# (``r.<f> = coalesce(row.<f>, r.<f>)``) so the two ingest paths converge and a
+# re-run carrying a null ordinal never clobbers a value an earlier load set. The
+# canonical edge key is ``hadith_number_in_book`` (mapped from ``row.hadith_number``,
+# the in-book ordinal — Book 1 Hadith 1), not the bare ``hadith_number`` (da#65).
 _APPEARS_IN_QUERY = """\
 UNWIND $batch AS row
 MATCH (h:Hadith {id: row.hadith_id})
 MATCH (c:Collection {id: row.collection_id})
-MERGE (h)-[:APPEARS_IN {
-    book_number: row.book_number,
-    chapter_number: row.chapter_number,
-    hadith_number: row.hadith_number
-}]->(c)
+MERGE (h)-[r:APPEARS_IN]->(c)
+SET r.book_number = coalesce(row.book_number, r.book_number),
+    r.chapter_number = coalesce(row.chapter_number, r.chapter_number),
+    r.hadith_number_in_book = coalesce(row.hadith_number, r.hadith_number_in_book)
 """
 
 _APPEARS_IN_CHECK = """\
--- a/tests/integration/test_graph_loading.py
+++ b/tests/integration/test_graph_loading.py
@@ -416,6 +416,63 @@ class TestEdgeLoading:
         rows = neo4j_client.execute_read("MATCH ()-[r:APPEARS_IN]->() RETURN count(r) AS cnt")
         assert rows[0]["cnt"] >= 0  # at least the edges were attempted
 
+    def test_appears_in_null_in_book_ordinal_does_not_abort_batch(
+        self, neo4j_client: Neo4jClient, tmp_path: Path
+    ) -> None:
+        """ingest-platform sibling of da#77: a null in-book ordinal must NOT abort
+        the APPEARS_IN write batch.
+
+        Scraped hadiths carry a null in-book ordinal until it is enriched. The
+        previous ``MERGE (h)-[:APPEARS_IN {hadith_number: row.hadith_number}]->(c)``
+        raised ``Cannot merge relationship using null property value`` and killed
+        the load for *every* hadith in the batch. The mock suite cannot enforce
+        this Neo4j rule, so it must be exercised against a real container.
+
+        Post-fix the positional props are SET after the MERGE under the canonical
+        key ``hadith_number_in_book`` (da#65), so a null ordinal yields an edge
+        with that key simply absent (Neo4j drops a property SET to null) instead
+        of aborting — and the non-null sibling in the same batch still loads.
+        """
+        staging, curated = _write_staging_data(tmp_path)
+        # The Collection node id must match what load_edges builds from
+        # ``{source_corpus}:{collection_name}`` — the corpus-prefixed staging form
+        # ("sunnah:bukhari" -> node "col:sunnah:bukhari").
+        _write_collections_parquet(
+            staging, [{**SAMPLE_COLLECTIONS[0], "collection_id": "sunnah:bukhari"}]
+        )
+        # One hadith with a null in-book ordinal (the da#77 trigger) alongside a
+        # normal one, both in the same collection / write batch.
+        _write_hadiths_parquet(
+            staging,
+            [
+                {**SAMPLE_HADITHS[0], "source_id": "bukhari:1", "hadith_number": None},
+                {**SAMPLE_HADITHS[1], "source_id": "bukhari:2", "hadith_number": 2},
+            ],
+        )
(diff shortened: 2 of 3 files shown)


parametrization:
## Review — null-safe APPEARS_IN MERGE (#84)

Reviewed at head `5dcd77f`. The fix is correct and well-scoped.

**The MERGE is now null-safe.** The relationship pattern is reduced to its true identity, `MERGE (h)-[r:APPEARS_IN]->(c)` — a hadith appears in a collection at most once — with no inline property map. The positional props (`book_number`, `chapter_number`, in-book ordinal) are SET after the MERGE, each guarded by `coalesce(row.<f>, r.<f>)`. Neo4j drops a property SET to null, so a scraped hadith with a null ordinal yields an edge with the key simply absent instead of raising `Cannot merge relationship using null property value` and aborting the whole write batch. This is the exact da#77/da#65 shape.

**No key drift.** The canonical key `hadith_number_in_book` (mapped from `row.hadith_number`, the in-book ordinal) matches the streaming path end-to-end — I confirmed `src/models/edges.py` (the `AppearsIn` model field), `workers/normalize/processor.py`, and `workers/ingest/schema.py` all emit `hadith_number_in_book`, and the streaming neo4j writer uses the same `MERGE ... [r:...] ... SET r.<f> = coalesce(row.props.<f>, r.<f>)` coalesce-preserve form (`tests/workers/test_ingest_neo4j.py`). The two ingest paths converge, and a re-run carrying a null ordinal will not clobber a value an earlier load set. The legacy bare `hadith_number` edge key is gone.

**The `_APPEARS_IN_CHECK` endpoint-existence query is untouched and correctly carries no property predicate**, so it does not reintroduce a null-unsafe path. The row builder still passes `hadith_number` as the row field name, consistent with `row.hadith_number` in the query.

**Test coverage is right.** The integration test `test_appears_in_null_in_book_ordinal_does_not_abort_batch` exercises a real `neo4j_client`, loads a null-ordinal row alongside a non-null sibling in the same batch, and asserts `appears_in.created == 2` (batch not aborted), the null edge has the key absent, and the sibling carries the ordinal — precisely what the mock historically masked. The two structural guard tests (`test_appears_in_merge_has_no_property_key_null_unsafe`, `test_appears_in_query_sets_ordinal_after_merge`) lock the MERGE line to its bare form and forbid the keyed regression. The mock test asserts a null ordinal is not skipped and still reaches the batch.

No `--no-verify`, no out-of-scope churn (3 files, +138/-6, all on the fix and its tests). The doc comment is accurate and earns its length.

Approving.

---
Requestor: Léopold Mbongo
Requestee: Imelda Santos
RequestOrReplied: Approved

TechDebt: none

parametrization:
## Second review — data-correctness lens (#85) — Approved

Reviewed at the PR head (5dcd77f), independent of the description framing.

### What is correct and well-built

- Moving the positional props out of the MERGE relationship pattern into a post-MERGE SET is the right shape. Identity of an APPEARS_IN edge is the (hadith, collection) pair; the in-book ordinal describes where in the collection and is not key material. Keying the MERGE on it was the abort cause and that is now gone — null-safe.
- The real-Neo4j integration test is genuine coverage, not mock-masked. It batches a null-ordinal row with a non-null sibling, asserts both edges created (batch not aborted), the null edge has the ordinal key absent (Neo4j drops a null SET rather than storing null), and the sibling carries `2` under the canonical key. That exercises the exact rule the mock cannot enforce.
- The mock structural guards (no inline property map on the MERGE line, canonical coalesce key present, bare legacy form absent) are a sound regression fence against reintroducing the null-unsafe shape.
- The coalesce-preserve SET is the correct non-clobber contract for a re-run carrying a null ordinal.

### On cross-path key convergence — verified, it holds

I checked the key-name claim across both ingest paths at the PR head, because cross-path consistency is the data-correctness crux here. They converge on the modeled name:

- the batch loader (this change) emits the Neo4j property `r.hadith_number_in_book`
- the streaming normalize path emits its APPEARS_IN edge payload under the same `hadith_number_in_book` key, mapped from the `hadith_number` source column
- the ingest edge allow-list carries only `hadith_number_in_book` for APPEARS_IN, the bare legacy key having been dropped per the #35 data-team ruling
- the AppearsIn model exposes `hadith_number_in_book`, so every producer matches the consumer

So a hadith loaded through either ingest path lands its in-book ordinal under one canonical key the API reads. That is exactly the convergence the change intends, and it is real at head.

For the record, an earlier draft of this review flagged a streaming-vs-batch divergence. That was my error — I had read the normalize processor and the edge allow-list from a stale working tree on a different branch, where those files still carried the bare legacy key. Re-read against the PR head sha, the convergence is correct and I withdraw that concern in full. The bug fix and the rename are both right.

### Scope

Three files, no `--no-verify`, fix matched to #84. Clean.

Approving — nice, well-tested fix.

---
Requestor: Camila Restrepo
Requestee: Imelda Santos
RequestOrReplied: Approved
TechDebt: none

PR #88, opened 2026-06-15
feat(metrics): pipeline metrics (Kafka lag, B2 size) + reprocess-from-stage (#76)

### LINKED ISSUE #76: Expose pipeline metrics (Kafka lag, B2 object-store size) + reprocess-from-stage for the admin data panel (relocated from isnad-graph#806)
## Context
Relocated from isnad-graph ig#806 (admin data-management panel) per the mis-filed-wave-issue rule — these AC need **ingest-platform/pipeline connectors that do not exist in isnad-graph**, so they belong here, not as isnad-graph debt.

## Scope (pipeline-side)
1. **Reprocess-from-stage** — a write op to re-run the pipeline from a given stage (raw/dedup/enriched/normalized/staged). Pairs with the existing reset surface (#9 / ingest#73).
2. **Pipeline metrics exposure** — Kafka consumer-group lag + B2 (object-store) size/usage, surfaced via an endpoint the isnad-graph admin data panel (or Grafana) can read. These need the Kafka admin client + B2 connector that live here.
3. **Grafana panels** (deploy/observability) chart the above once exposed — a downstream sub-task for the deploy repo once the metrics endpoint exists; note as a dependency, file in deploy when ready.

## Refs
isnad-graph ig#806, PR #986 (read-only inventory shipped — these write/infra features deferred). ingest#73 (reset endpoints, the sibling surface).

parametrization:
## Summary

Closes #76. Adds the admin-only HTTP surface the isnad-graph admin data panel consumes to monitor and steer the pipeline, plus a non-destructive reprocess-from-stage capability.

All routes mount under `/admin` and are guarded by `require_admin` (same user-service RS256 admin JWT as the existing reset endpoints) — no unauthenticated path to any operation.

### Endpoints
- **`GET /admin/metrics/lag`** — per-stage, per-partition Kafka consumer lag + pipeline total. `lag = end_offset − committed_offset` (clamped at 0; an uncommitted partition counts its full head as backlog).
- **`GET /admin/metrics/storage`** — per-B2-prefix object count + bytes, plus a bucket-wide rollup, over the `raw/ dedup/ enriched/ normalized/ staged/` stage layout.
- **`POST /admin/reprocess`** — re-run the pipeline from a chosen stage **without deleting staged B2 data** (that's the reset surface). Rewinds the entry stage's consumer group to the start of its consume topic and clears the idempotency checkpoints for that stage and every downstream stage, so re-delivered batches are reprocessed rather than skipped as already-seen. `dry_run` previews the plan and writes a dry-run audit entry without building a live client (mirrors reset).

### Design
- Core logic in `src/pipeline/metrics.py` and `src/pipeline/reprocess.py` is pure over small injected Protocols (`KafkaOffsetReader`, `ObjectSizeStore`, `OffsetRewinder`, `CheckpointResetter`), so the unit suite exercises the lag arithmetic, prefix-summing pagination, and rewind/checkpoint-clear orchestration with in-memory fakes — no Kafka, S3/B2, or Postgres. Live adapters (`*_adapters.py`) keep SDK imports function-local, matching the reset surface.
- Canonical stage topology in `src/pipeline/stages.py` is re-declared in `src` (rather than importing `workers`, preserving the `workers → src` direction) and **pinned** to `workers.lib.topics` + the worker entrypoints by `tests/test_pipeline/test_stages.py`, so drift fails a test instead of silently mis-reading a stage's lag.
- Exported `CHECKPOINT_TABLE` from `workers.lib.checkpoint_pg` (and routed the DDL through it) so the reprocess checkpoint-resetter clears the exact table the workers create — pinned by test.
- Every reprocess (dry-run included) writes an audit entry under `data/audit/`, consistent with the reset surface.

### Contract / docs
The metric + endpoint contract for the admin-panel consumer is documented in `docs/pipeline-metrics-api.md` (stage vocabulary, response schemas, dry-run semantics, and the consumer-group-rebalance operational note).

## Test Plan
- `ruff check` + `ruff format --check` over `src/ workers/ tests/` — clean.
- `mypy src/` (strict) — clean.
- `pytest -m "not integration"` — **681 passed, 4 skipped** (includes 36 new tests across stages/metrics/reprocess core + both API routers).
- `markdownlint-cli2` over the docs corpus — 0 errors (new doc included).

### What can't be unit-tested here (runtime acceptance)
The live Kafka offset reads, B2 listing, and the actual offset-rewind/checkpoint-clear against a running broker + Postgres require a deployed stack. The adapter mapping logic is covered with fake admin/consumer objects; end-to-end behavior is a post-merge staging check. Per the operational note in the doc, an offset rewind only takes effect for a consumer group with no active members on the partition (same contract the reset endpoints already carry).

## Reviewers
@bjornhenriksen @petravidovic — please focus on: (1) the reprocess semantics (clear-before-rewind ordering; clearing downstream checkpoints vs. rewinding only the entry group); (2) the lag arithmetic edge cases (uncommitted partition, committed-ahead clamp); (3) the src↔workers stage-table pinning approach.

### DIFF
--- a/docs/pipeline-metrics-api.md
+++ b/docs/pipeline-metrics-api.md
@@ -1,0 +1,134 @@
+# Pipeline metrics & reprocess API (issue #76)
+
+Admin-only HTTP surface the **isnad-graph admin data panel** fetches to show
+pipeline health and to trigger a reprocess. All routes are served by the
+ingest-platform FastAPI app (`src/api/app.py`), mounted under `/admin`, and
+guarded by `require_admin` — the same user-service RS256 admin JWT that
+authorizes the isnad-graph admin dashboard and the reset endpoints.
+
+- Base path: `/admin`
+- Auth: `Authorization: Bearer <user-service access token with role admin>`
+  - missing/invalid token → `401`; valid non-admin → `403`; user-service
+    JWKS unreachable → `503`
+
+Run locally: `uvicorn src.api.app:create_app --factory --port 8002`.
+
+## Stage vocabulary
+
+The pipeline runs four consuming stages, in order. The `stage` names below
+are the stable identifiers used by both the metrics and reprocess routes.
+
+| stage | consumer group | consumes topic | produces topic |
+| --- | --- | --- | --- |
+| `dedup` | `dedup-worker` | `pipeline.raw.landed` | `pipeline.dedup.done` |
+| `enrich` | `enrich-worker` | `pipeline.dedup.done` | `pipeline.enrich.done` |
+| `normalize` | `normalize-worker` | `pipeline.enrich.done` | `pipeline.normalize.done` |
+| `graph-load` | `ingest-worker` | `pipeline.normalize.done` | (terminal) |
+
+This table is the source of truth in `src/pipeline/stages.py` and is pinned
+to the worker wiring by `tests/test_pipeline/test_stages.py`.
+
+## GET /admin/metrics/lag
+
+Per-stage Kafka consumer lag — how far each stage's consumer group is behind
+the head of the topic it consumes. `lag = end_offset - committed_offset`,
+clamped at 0; a partition the group has never committed counts its whole
+head offset as backlog. `total_lag` (per stage and overall) is the sum over
+partitions/stages — the headline "is the pipeline keeping up?" number.
+
+Response `200`:
+
+```json
+{
+  "stages": [
+    {
+      "stage": "dedup",
+      "consumer_group": "dedup-worker",
+      "topic": "pipeline.raw.landed",
+      "total_lag": 12,
+      "partitions": [
+        { "partition": 0, "end_offset": 100, "committed_offset": 88, "lag": 12 }
+      ]
+    }
+  ],
+  "total_lag": 12
+}
+```
+
+- `committed_offset` is `null` when the group has never committed that
+  partition (the partition's `lag` then equals `end_offset`).
+- `stages` always lists all four stages in pipeline order; a stage whose
+  topic has no partitions yet reports `total_lag: 0` and `partitions: []`.
+
+## GET /admin/metrics/storage
+
+B2 object-store size, broken down by the pipeline's stage prefixes plus a
+bucket-wide rollup. `total_bytes` is the sum of object sizes under each
+prefix; `object_count` is the number of objects.
+
+Response `200`:
+
+```json
+{
+  "prefixes": [
+    { "prefix": "raw/", "object_count": 1432, "total_bytes": 9105533 },
+    { "prefix": "dedup/", "object_count": 1280, "total_bytes": 8401200 },
+    { "prefix": "enriched/", "object_count": 1280, "total_bytes": 9550120 },
+    { "prefix": "normalized/", "object_count": 640, "total_bytes": 4120044 },
+    { "prefix": "staged/", "object_count": 640, "total_bytes": 4010990 }
+  ],
+  "total_object_count": 5272,
+  "total_bytes": 35187887
+}
+```
+
+The prefixes match the reset surface's B2 stage layout (`raw/`, `dedup/`,
+`enriched/`, `normalized/`, `staged/`).
+
+## POST /admin/reprocess
+
+Re-run the pipeline from a chosen stage **without deleting staged B2 data**
+(that is the reset surface, `POST /admin/reset/*`). Reprocessing from stage
+*X* rewinds *X*'s consumer group to the start of its consume topic and clears
+the idempotency checkpoints for *X* and every downstream stage, so the
+re-delivered batches are processed again instead of being skipped as
+already-seen. Only the entry stage's group is rewound — downstream stages
+re-consume naturally as *X* re-emits.
+
+Request:
+
+```json
+{ "stage": "enrich", "dry_run": false }
+```
+
+- `stage` (required) — one of the stage names above. Unknown → `422`.
+- `dry_run` (default `false`) — resolve the plan and write a dry-run audit
+  entry **without** touching Kafka or Postgres. The admin panel should call
+  this first to preview the affected stages.
+
+Response `200`:
+
+```json
+{
+  "from_stage": "enrich",
+  "dry_run": false,
+  "audit_entry_path": "data/audit/2026-06-14T….json",
+  "summary": {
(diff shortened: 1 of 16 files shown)


parametrization:
Requestor: Bjørn Henriksen
Requestee: Fatima Bensalah
RequestOrReplied: Approved
TechDebt: #91

**Review: metrics + reprocess-from-stage are correct, well-isolated, and thoroughly tested — ship it**

Reviewed at HEAD (971e20d), all 16 files + the three named focus areas. Integration points cross-checked against HEAD: `create_audit_entry`/`write_audit_entry` signatures, `KafkaPythonAdmin.reset_consumer_offsets`, and `reset.STAGE_PREFIXES` all match how this PR consumes them.

Focus 1 — reprocess ordering. Clear-before-rewind is right, and the test pins the exact event order (`clear:normalize-worker`, `clear:ingest-worker`, `rewind:normalize-worker`). Clearing first closes the window where a re-delivered batch could land and be skipped by a stale checkpoint before the clear runs. Clear-downstream-too is also right — only the entry group is rewound, downstream stages re-consume naturally at the tail of their topics as X re-emits, but their `(stage, batch_id)` checkpoints carry the original batch_id, so without clearing them every downstream stage would treat the re-delivery as already-done and the reprocess would be a silent no-op. `downstream_stages` correctly returns entry+successors in order; terminal-stage case touches only itself. Sound.

Focus 2 — lag edge cases. `baseline = committed if not None else 0; lag = max(0, end - baseline)` handles both the never-committed partition (full head as backlog) and committed-ahead (clamped to 0) exactly as documented, each with a dedicated test. The adapter also drops kafka-python's `-1` sentinel so a never-committed partition reads as absent rather than a bogus offset. The full-backlog-on-never-committed convention assumes earliest-reset consumption — a documented, conservative choice for an "is it keeping up" signal, fine as-is.

Focus 3 — stage-table pinning. Re-declaring the topology in `src` (not importing `workers.lib.topics`) keeps the src→workers dependency direction clean and matches the reset.py precedent. Topics and the checkpoint-table name are pinned against the real worker constants (`workers.lib.topics`, `workers.lib.checkpoint_pg.CHECKPOINT_TABLE`), so those cannot drift silently. See the tracked tech-debt note below on the consumer-group pin.

Auth/scope is consistent with the reset surface — every router mounted under `/admin` with `require_admin` at the mount point, real-guard 401 tests on both surfaces, dry-run never constructs a live client (factory spy `build_count == 0`), unknown stage rejected as 422 before any build. Runtime acceptance (live offsets / B2 / actual rewind) deferred to post-merge staging is acceptable for this layer.

CI: the lone red check (`precommit-ci-sync` DRIFT build) is the pre-existing main-branch false-fail tracked in ingest#89, not introduced here — not held against this PR.

No must-fix items.

Tech-debt finding, now filed as #91 — the consumer-group assertion in test_stages.py pins the stages.py declaration against a hardcoded literal dict, not against the actual `workers.<stage>.main` WorkerSettings, even though the module docstring states the ids "match workers.<stage>.main's WorkerSettings." Topics are pinned against real worker constants; consumer-group ids are not. A worker entrypoint renaming its consumer_group would clear the wrong checkpoint rows on reprocess without failing a test. Tracked in #91 for a follow-up that imports the worker settings to pin the group ids the same way the topics are pinned.

parametrization:
Requestor: Petra Vidović
Requestee: Fatima Bensalah
RequestOrReplied: Approved
TechDebt: #92, #93

**Review: Clean Protocol-injected metrics + reprocess surface; consumer contract documented and pinned, dry-run proven inert, admin-gated at the mount — approve.**

Reviewed at HEAD (`F.Bensalah/0076-pipeline-metrics`), not PR-body framing. Walked the implementation and every test.

Consumer-facing contract (isnad-graph admin panel) — solid. The three shapes (lag, storage, reprocess) are frozen pydantic models that match `docs/pipeline-metrics-api.md` field-for-field; never-committed partition maps to a null `committed_offset` with lag equal to head, and all four stages always list in order — exercised by `test_lag_response_shape` / `test_storage_response_shape`. The stage vocabulary is the single source of truth in `stages.py` and is pinned to `workers.lib.topics` plus the worker entrypoints by `test_stages.py`, so the wire-level topic/group strings cannot silently drift from what the workers actually read.

Reprocess correctness — the clear-before-rewind ordering is the right call and is locked by `test_reprocess_clears_checkpoints_before_rewinding` (asserts the exact event sequence), so a re-delivered batch can never hit a stale checkpoint between the two operations. Only the entry stage's group is rewound; downstream re-consume naturally — matches the docstring and the audit summary. `audit_entry_path` + `rows_affected` (sum of cleared counts) are correct, and the checkpoint key being the `consumer_group` is documented and pinned (`_CHECKPOINT_TABLE` pinned to `workers.lib.checkpoint_pg.CHECKPOINT_TABLE`; SQL uses a fixed-constant table name + bound stage param, no injection surface).

dry-run inertness — proven, not asserted. `factory_spy.build_count == 0` and `fake_reprocessor.calls == []` for the dry-run path; the factory dependency returns a closure so resolving it opens no Kafka/PG connection, and the live client is built only in the non-dry-run branch. Unknown stage is a 422 before any client is built.

Admin gating — both metric routes and the reprocess route are guarded once at the `/admin` mount with `Depends(require_admin)`; the auth-gate tests use the REAL guard and confirm 401 on a missing header. The Protocol-over-adapters split keeps `src` independent of `workers` and gives full unit coverage with no broker/S3/PG.

Runtime acceptance (live broker rewind, offset-0-vs-earliest under Kafka retention, the no-active-member offset-alter constraint, real B2 listing) is correctly deferred to post-merge staging — acceptable.

Not withholding over the red `precommit-ci-sync` check — pre-existing publish-workflow false-fail on main since #87, tracked as ingest#89, not this PR's doing.

Must-fix: None

Non-blocking findings, now filed as tech-debt issues (see the TechDebt line in the header block):

- #92 — `STORAGE_PREFIXES` and the reprocess table name are re-declared in `src` and kept honest only by test pins. Intentional to preserve the `src ↛ workers` import direction, fine as-is, but worth revisiting if a fourth consumer of these constants ever appears.
- #93 — the dry-run route path resolves `plan_reprocess` twice (once for the up-front 422 guard, again inside `write_dry_run_audit`). Harmless (pure planning call, no connection opened), trivially de-dupable later.

PR #115, opened 2026-06-23
ci(base-pin): wire Dockerfile base-pin gate into CI/pre-commit + README local-hooks docs (#114)

### LINKED ISSUE #114: Roll base-pin check into CI/pre-commit + README local-hooks docs (parent main#744/#718)
Roll the charter-prose→code base-pin gate (built in noorinalabs-main#735) into this repo's CI + pre-commit + sync-drift classifier, and document the local hooks in the README. The ingest-platform slice of the cross-repo rollout.

## Scope

- **#744 (base-pin):** wire `.claude/lib/check_dockerfile_base_pin.py` over the five worker images (top-level multi-worker `Dockerfile` + `workers/{dedup,enrich,normalize,ingest}/Dockerfile`) into:
  - `.github/workflows/ci.yml` (new `base-pin` job)
  - `.pre-commit-config.yaml` (new commit-stage `dockerfile-base-pin` hook)
  - `.claude/lib/pre_commit_ci_sync.py` (classify the `dockerfile-base-pin` kind so the local mirror is mandatory — #684 parity)
  - fix-forward the violations the first run surfaces.
- **#718 (docs):** add/refresh the README "Local hooks" section, mirroring noorinalabs-main CLAUDE.md § Local Hooks (install both stages, the full commit/push check-set, the sync-drift gate).

## fixture-realism — not applicable here

`check_fixture_realism.py` targets Arabic NER / graph-load **data fixtures**. This repo carries none: Arabic text appears only inline in `tests/test_parse/*.py` unit sources, and there is no curated-fixture directory. So this slice scopes to base-pin + the README, and intentionally does NOT wire a fixture-realism gate (nor classify that kind).

## Provenance

- Parent meta: noorinalabs-main#744 (roll base-pin + fixture-realism checks out to child-repo CI, from #735)
- Parent meta: noorinalabs-main#718 (README local-hooks docs)

parametrization:
## Summary

Ingest-platform slice of the cross-repo base-pin + local-hooks-docs rollout
(parent meta noorinalabs-main#744 + #718, gates built in #735). Wires the
charter-prose→code **base-image-pin** gate into this repo's CI, pre-commit, and
sync-drift classifier; fix-forwards the Dockerfiles it flags; and refreshes the
README local-hooks docs.

Closes #114.

## What changed

**#744 — base-pin gate**
- Vendored `.claude/lib/check_dockerfile_base_pin.py` (+ its unit tests) from the
  parent template.
- `.github/workflows/ci.yml`: new `base-pin` job running the check over all five
  worker images (top-level multi-worker `Dockerfile` + `workers/{dedup,enrich,
  ingest,normalize}/Dockerfile`). The `precommit-ci-sync` job now runs the whole
  `.claude/lib/tests/` dir so the new check's tests gate the PR.
- `.pre-commit-config.yaml`: new commit-stage `dockerfile-base-pin` hook over the
  same five files.
- `.claude/lib/pre_commit_ci_sync.py`: classify the `dockerfile-base-pin` kind so
  a CI gate with no local mirror is harmful drift, not silence (#684 parity).
- **Fix-forward** the 5 Dockerfiles the first run flagged (10 violations):
  digest-pin `python:3.14-slim@sha256:44dd…50061` (digest kept in lockstep with
  the sibling repos' pin) and add `apt-get -y upgrade`.

**#718 — README local-hooks docs**
- Refreshed the README "Local hooks" section, mirroring noorinalabs-main
  CLAUDE.md § Local Hooks: both install stages, the full commit/push check-set
  (now including cspell + `dockerfile-base-pin`), and the sync-drift gate.

## fixture-realism — not applicable

`check_fixture_realism.py` was **not** wired and its kind was **not** classified:
this repo carries no Arabic NER/graph-load **data** fixtures — Arabic appears
only inline in `tests/test_parse/*.py` unit sources, and there is no curated-
fixture directory. The slice therefore scopes to base-pin + docs.

## Evidence (local, all green)

- `check_dockerfile_base_pin.py` over the 5 Dockerfiles: `OK` (was 10 violations
  before the fix-forward).
- `pre_commit_ci_sync.py .`: `OK: pre-commit config mirrors all CI-enforced checks.`
- `pytest .claude/lib/tests/`: 52 passed.
- `pre-commit run --hook-stage pre-commit`: actionlint + dockerfile-base-pin Passed.
- `pre-commit run --hook-stage pre-push`: ruff, actionlint, cspell, pip-audit,
  mypy-strict, pytest all Passed.
- `markdownlint-cli2`: 0 errors.

## Reviewers

- Peer reviewer: **Petra Vidović** (tech_lead_petra)
- Secondary reviewer: **Fatima Bensalah** (devops_engineer_fatima)

## Provenance

- noorinalabs-main#744 — roll base-pin + fixture-realism checks out to child-repo CI (from #735)
- noorinalabs-main#718 — README local-hooks docs

https://claude.ai/code/session_01W7PudS35xmg2FW6tJekTGP

### DIFF
--- a/.claude/lib/check_dockerfile_base_pin.py
+++ b/.claude/lib/check_dockerfile_base_pin.py
@@ -1,0 +1,217 @@
+#!/usr/bin/env python3
+"""Lint Dockerfile `FROM` statements for digest-pin + distro-upgrade compliance.
+
+Deterministic conversion of `.claude/team/charter/tech-decisions.md § Base Image
+Pinning` (noorinalabs-main#735, the charter-prose-inventory worklist item #4 /
+epic #726). The section requires every `FROM` to use a **digest-pinned tag**
+(`image:tag@sha256:<digest>`) **combined with an in-image package upgrade**, to
+close the two failure modes it names — floating-tag drift and within-tag package
+drift (the shape isnad-graph#853 hit).
+
+What this gate enforces (verbatim from the section's distro table)
+================================================================
+For every `FROM` that is not exempt:
+
+1. The image reference MUST carry an `@sha256:` digest pin.
+2. The stage MUST run the package-manager upgrade matching the base distro:
+
+   | Family      | Pin shape               | Upgrade command                                 |
+   |-------------|-------------------------|-------------------------------------------------|
+   | Alpine      | image:tag@sha256:digest | `RUN apk upgrade --no-cache`                    |
+   | Debian-slim | image:tag@sha256:digest | `apt-get update && apt-get -y upgrade && clean` |
+   | Distroless  | image:tag@sha256:digest | none — no package manager (pinned-only is fine) |
+
+   The distro family is inferred from the image name: `alpine` → Alpine,
+   `distroless` → Distroless (no upgrade required), otherwise the glibc/Debian
+   default (an apt upgrade is required). A genuinely-other base that has no
+   package manager should carry the `# RATIONALE:` exemption below.
+
+Exemptions honored (verbatim from the section)
+=============================================
+- **`scratch`** final/any layer — no package manager; the upstream stages that
+  produced its contents still follow the rule and are checked on their own
+  `FROM`.
+- **Distroless** — pinned-only is sufficient (no package manager); the digest
+  pin is still required.
+- **Multi-stage stage reference** — a `FROM <earlier-stage>` inherits the
+  upstream stage's pin, so it is not re-checked (the named stage was checked at
+  its defining `FROM`).
+- **Vendor images** documented with an inline `# RATIONALE:` comment on the
+  `FROM` line (or the comment line immediately above it).
+
+CLI
+===
+    python3 .claude/lib/check_dockerfile_base_pin.py <Dockerfile> [<Dockerfile> ...]
+
+Exit codes:
+    0 — every FROM in every file is compliant (or exempt)
+    1 — at least one violation (each printed as `path:line: FROM <image> — <why>`)
+    2 — usage / file-not-found error
+
+This is a reusable template (same CLI/exit-code shape as
+`.claude/lib/pre_commit_ci_sync.py`) so it wires identically into a repo's
+pre-commit config and its CI. The parent repo `noorinalabs-main` ships no
+Dockerfiles itself; the cross-repo rollout that points this at each child's
+Dockerfiles is a #735 follow-up.
+"""
+
+from __future__ import annotations
+
+import re
+import sys
+from pathlib import Path
+
+# `FROM` is the only keyword we parse; Dockerfile keywords are case-insensitive.
+_FROM_RE = re.compile(r"^\s*FROM\s+(?P<rest>\S.*?)\s*$", re.IGNORECASE)
+# A leading `--platform=...` flag precedes the image and must be stripped.
+_PLATFORM_RE = re.compile(r"^--platform=\S+\s+", re.IGNORECASE)
+# A trailing `AS <stage>` names the build stage.
+_AS_RE = re.compile(r"\s+AS\s+(?P<stage>\S+)\s*$", re.IGNORECASE)
+
+# Upgrade-command signatures per distro family (searched over the stage body).
+_ALPINE_UPGRADE_RE = re.compile(r"\bapk\s+upgrade\b", re.IGNORECASE)
+# `apt-get update && apt-get -y upgrade` / `apt upgrade` — an apt invocation
+# followed (in the same command segment) by `upgrade`. `apt-get update` alone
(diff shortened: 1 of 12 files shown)


parametrization:
Requestor: Petra Vidović
Requestee: Bjørn Henriksen
RequestOrReplied: Approved
TechDebt: none

Reviewed at head 133058650833ece35e83a6f2ea0990f135a2c3f5 against the diff and the files at that ref — not the PR body.

The gate itself is sound. `check_dockerfile_base_pin.py` parses every `FROM` correctly: it strips a leading `--platform=` flag and a trailing `AS <stage>`, drops inline comments, records defined stage names so a later `FROM <stage>` inherits the upstream pin and is not re-checked, and exempts `scratch`, distroless (pin-only), and a documented `# RATIONALE:` (inline or on the comment line immediately above). The distro split is right — Alpine wants `apk upgrade`, everything non-distroless/non-scratch wants an apt upgrade, and the apt regex deliberately does not accept a bare `apt-get update` as satisfying the upgrade floor (the `test_apt_update_only_is_not_an_upgrade` case nails that). The test module is genuinely adversarial rather than a mirror of the implementation: positive cases use the required voweled-shape pattern, negatives use the exact prohibited forms (floating tag, pin-only, apk-only, update-only), and the multi-stage mixed case confirms only the offending runtime stage is flagged. Good work.

The five Dockerfiles are all converted consistently — same digest across the top-level multi-worker image and the four worker stages, with `apt-get -y upgrade` spliced into the existing RUN chain before the install step. I confirmed the pinned manifest resolves on the registry, so the digest is real and not a transcription error.

On the fixture-realism not-applicable determination: the conclusion is correct — there is no curated Arabic data-fixture directory for the gate's `files:` discovery to point at (the `data/` tree is `.gitkeep` placeholders only), and aiming the lens at Python unit-test sources would false-flag the deliberately-degenerate Arabic in the negative-path parser tests. So scoping this slice to base-pin + docs is the right call. One precision point for the record, since it is stated as the justifying rationale in both the PR body and the sync-classifier comment: Arabic inline strings are not confined to `tests/test_parse/*.py` — they also appear in `tests/conftest.py`, `tests/factories.py`, the integration tests, and several other `tests/test_*` modules. That does not change the outcome (none of these are curated fixtures the lens is meant to judge), but the narrower phrasing slightly undersells the spread. Worth tightening if the comment is ever lifted into a sibling repo's slice.

No blockers. Approved.

parametrization:
Requestor: Fatima Bensalah
Requestee: Bjørn Henriksen
RequestOrReplied: Approved
TechDebt: sibling python:3.14-slim pins (isnad-graph, data-acquisition) not yet on this digest — #744 rollout should converge them.

Reviewed at head 133058650833ece35e83a6f2ea0990f135a2c3f5, checking the CI / pre-commit / sync wiring against the files at that ref.

The three-surface wiring is coherent and mutually reinforcing:

- The new CI job runs the stdlib-only checker over all five Dockerfiles as a single-line `run:`, which is exactly what the line-scanning sync classifier needs to register the script name — so the job advertises the kind it belongs to.
- The pre-commit hook lists the same five files in its `entry` with `pass_filenames: false`, while the `files:` regex scopes when it triggers. That means a developer's commit and a CI run check the identical set, which is the property that keeps the two from drifting on which images are covered.
- The classifier gains a `dockerfile-base-pin` kind with both a CI-step pattern and a pre-commit-id pattern, and the added drift tests prove it both ways: mirror present is not harmful, CI-only is harmful. So dropping either side now fails the sync gate rather than going silent — the #684 parity stance, applied correctly.

Broadening the sync job from the single sync-test file to the whole `.claude/lib/tests/` directory is necessary, not cosmetic: `testpaths = ["tests"]` in pyproject means the main test job never collects the gate tests under `.claude/lib/tests/`, so without this change the new checker's tests would gate nothing. Good catch.

I confirmed the pinned manifest resolves on Docker Hub, so the digest is valid — worth stressing because the static gate only checks for the presence of an `@sha256:` marker, not that it points at a real image; a bad digest would slip past this job and only blow up at image-build time.

One operational item, tracked above as tech debt rather than a blocker for this slice: the comments assert the digest is "kept in lockstep with the sibling repos' python:3.14-slim pin", but isnad-graph and data-acquisition still carry an unpinned `FROM python:3.14-slim` in their trees — their #744 slices are presumably in flight. So the lockstep is aspirational right now. The cross-repo rollout should make sure they all land on this same digest; if they diverge, the "lockstep" comments become misleading. Not this PR's job to fix, but flagging it for the #744 coordinator.

README accurately reflects both install stages, the full commit/push check-set including `dockerfile-base-pin`, and the sync-drift gate. Approved.

Issue #119, opened 2026-06-26
grade_normalized parity: mirror normalize_grade() in the streaming ingest worker — da#153 item #4


parametrization:
**Owner decision (2026-06-26, P7W20 retro of da#153): file the cross-repo parity follow-up.**

`noorinalabs-data-acquisition` PR #150 added `normalize_grade()` + the `grade_normalized` property on the **batch** load path (`src/graph/load_nodes.py`). The Kafka streaming worker here (`workers/ingest`) must mirror the same normalization so the batch and streaming paths converge on identical `grade_normalized` values.

**Work:** apply the same `normalize_grade()` logic in the streaming ingest worker. This is the parity change; the data-acquisition batch path is already correct.

Provenance: cross-repo follow-up from noorinalabs-data-acquisition#153 / ADR-004 item #4. Meta: noorinalabs-main#723.

parametrization:
Resolved by PR #121 (merged to the P8W24 wave branch `deployments/phase-8/wave-24`). `grade_normalized` now mirrors the canonical `normalize_grade()` in the streaming normalize worker on both the Hadith and Grading nodes, with parity tests pinning streaming == batch == canonical. Reviewed + approved by Petra Vidović and Tomás Carvalho; drift-safe once the ig sibling (ig#1168) landed. Closing manually — wave-branch merges do not auto-close.

PR #121, opened 2026-07-06
tech-debt(ip#119): grade_normalized parity — mirror normalize_grade() in the streaming path

parametrization:
Part of #119 (da#153 ADR-004 item #4; meta noorinalabs-main#723).

## Problem — parity gap was REAL at wave-branch HEAD
The batch load path (`noorinalabs-data-acquisition/src/graph/load_nodes.py`, added by da#150) writes `grade_normalized = normalize_grade(grade)` on **both** the Hadith node and the Grading node. The Kafka streaming path did **not**: `workers/normalize/processor.py` emitted only the raw `grade`, and `workers/ingest/schema.py`'s allow-list did not permit `grade_normalized` — so a streamed load and a batch load of the same hadith diverged on that field.

## Fix — close it end-to-end in the streaming path
- **`src/utils/grade.py`** (new) — verbatim mirror of the batch `normalize_grade()`, wired to this repo's own `HadithGrade` enum and `normalize_arabic`. Verified value-equivalent to data-acquisition's: both `normalize_arabic` implementations produce identical output on every grade needle (صحيح/حسن/ضعيف/موضوع/li-ghayrihi/voweled forms), and the two `HadithGrade` enums are byte-identical.
- **`workers/normalize/processor.py`** — emit `grade_normalized` on the Hadith node (always — an absent grade normalizes to `"unknown"`, matching the batch node) and on the Grading node (only when a grade is present).
- **`workers/ingest/schema.py`** — allow-list `grade_normalized` on `Hadith` + `Grading` so the emitted value survives ingest instead of being silently dropped.
- Regenerated `ontology/structural/` for the new module.

## Tests (workers suite, no Docker: `ENVIRONMENT=test make test-workers`)
- **`test_grade_normalized_mirrors_batch_normalize_grade`** — proves parity: for English / bare Arabic / voweled Arabic / li-ghayrihi qualifier / unrecognised / missing grade, the streaming `grade_normalized` equals both `normalize_grade(raw)` and a pinned canonical value, on Hadith and Grading nodes.
- **`test_grade_normalized_is_landing_eligible_at_ingest`** — guards the ingest allow-list so a future edit can't silently re-open the gap.
- Extended `test_fan_out_produces_expected_per_label_counts`.

## Cross-repo follow-up REQUIRED (does not block this PR's CI)
`grade_normalized` is ingest-only, exactly like `grade` already is (the isnad-graph model exposes only `grade_composite`). isnad-graph's `schema-drift.yml` gate will flag `grade_normalized` as a phantom ingest field on its next PR in this wave window until a sibling entry is added to `noorinalabs-isnad-graph/scripts/ingest-extras.yaml` under `nodes.Hadith` **and** `nodes.Grading` (category `permanent`, anchored to ADR-005), mirroring the existing `grade` entry. This is a designed convergence window (the gate resolves to the matching wave branch); flagging for sequencing.

TechDebt: none


parametrization:
Requestor: Tomás Carvalho
Requestee: Imelda Santos
RequestOrReplied: Approved

**Review: Tests genuinely prove batch⇄streaming grade_normalized parity across the requested matrix on both node labels; full workers suite green at HEAD.**
Must-fix: None
TechDebt: No concrete pinned case for the HASAN / HASAN_LI_GHAYRIHI / MAWDU enum values (incl. the mawdu English "fabricat" needle). Every rule *shape* is exercised (bare grade, li-ghayrihi ordering, Arabic normalization, English synonym, unknown fallback, missing) so parity risk is low — worth a follow-up to pin the hasan-family/mawdu rows.

QA notes (not blockers):
- Matrix coverage confirmed: English (`sahih`, `da'if`→daif), bare Arabic (`صحيح`, `ضعيف`), voweled Arabic (`صَحِيحٌ`), li-ghayrihi qualifier (`صحيح لغيره`→sahih_li_ghayrihi, ordered before bare sahih), unrecognised (`totally made up`→unknown), and missing (`None`→unknown).
- Both labels covered: each case asserts the Hadith node's grade_normalized; when a grade is present it also asserts the Grading node's value, and asserts NO Grading node when grade is absent.
- Strong assertion form: `streaming == normalize_grade(raw) == pinned canonical value` — the pinned leg is the real parity anchor (a true cross-repo compare against da's function isn't feasible in this repo; pinning is the correct substitute) and catches drift in either copy. Verified pinned values against HadithGrade at HEAD.
- Ingest allow-list is guarded: `test_grade_normalized_is_landing_eligible_at_ingest` asserts the key is present in NODE_PROPERTY_MAP for both Hadith and Grading, so a future schema tightening can't silently re-open the gap.
- Ran the suite: `ENVIRONMENT=test` pytest in a detached worktree at head 7f2463c — 3 targeted tests pass; full `tests/workers/` = 93 passed, 1 skipped.
- isnad-graph schema-drift convergence (ig#1168) noted as out-of-scope, not flagged.

parametrization:
Requestor: Petra Vidović
Requestee: Imelda Santos
RequestOrReplied: Approved

**Review: grade.py is a verified value-equivalent mirror of the canonical normalize_grade — parity holds on every recognized grade form.**
Must-fix: None
TechDebt: ip src/utils/arabic.py has not absorbed da#271 strip_format_marks, so a raw grade carrying an embedded (mid-token) bidi/zero-width mark diverges (batch normalizes to the real grade, stream falls to unknown); recommend a follow-up to port strip_format_marks into the streaming normalize path (also benefits narrator names).

Verification (executed both normalize_arabic impls against the grade needles and inputs, not read from PR body):
- Needles byte-identical across both repos for all six families (sahih li-ghayrihi / hasan li-ghayrihi / mawdu / daif / sahih / hasan). The grade.py code body is byte-identical to noorinalabs-data-acquisition/src/utils/grade.py; only the docstring adds an ip#119 parity note.
- Both HadithGrade StrEnums have identical members and values (7 each).
- Every recognized form converges: English (sahih/weak/authentic/good/mawdu/fabricated), bare Arabic, voweled Arabic, the li-ghayrihi qualifier beating bare sahih, an unrecognised value, and a missing grade all yield the same grade_normalized as the batch path. Absent-grade to unknown matches; qualifier precedence matches.
- Ingest allow-list confirmed: grade_normalized is added to both the Hadith block and the Grading block of NODE_PROPERTY_MAP, so the value is not silently dropped at ingest; the landing-eligibility test pins both labels.
- The isnad-graph schema-drift (grade_normalized is ingest-only) is tracked separately as ig#1168 — expected convergence, not flagged here.

The only divergence between the two paths is the embedded-format-mark edge above, which stems from a pre-existing difference in the shared normalize_arabic dependency, not from the grade mirror itself. Substring matching already tolerates leading and trailing marks (verified identical). Grade is a small controlled-vocabulary field, so the embedded-mark case is low-probability; tracking it as TechDebt rather than blocking the parity close.

Issue #127, opened 2026-07-11
ci(publish): GHCR publish red on main since 07-06 — cached apt layer freezes 10 HIGH OS CVEs (repo-side of main#947)


parametrization:
## Summary

`Publish to GHCR` has been **red on `main` since 2026-07-06** ([run 28766990515](https://github.com/noorinalabs/noorinalabs-isnad-ingest-platform/actions/runs/28766990515)) — five days, undetected. Trivy fails on HIGH OS-package CVEs in the published image.

This is the repo-side half of **noorinalabs-main#947**: the buildx GHA layer cache serves the `RUN apt-get update && apt-get -y upgrade && ...` layer from cache, so the upgrade never executes and the packages never move.

## Measured evidence

I re-ran the failed run today (job `86502590623`):

- **`#11 CACHED`** on the apt layer — imported from `gha:2684583647406898998`. The upgrade did not run.
- Package versions **identical** to the 2026-07-06 build despite published fixes: `libssh2-1t64` still `1.11.1-1` (fixed: `1.11.1-1+deb13u1`), `linux-libc-dev` still `6.12.94-1` (fixed: `6.12.95-1`).
- HIGH CVE count **grew 5 → 10** between the two runs. A build that actually re-ran apt would have gone down. All 10 are `fixed` upstream.

## Why this repo is the one that went red first

Its cache entry is the oldest, not its Dockerfile the worst. Three sibling repos (`data-acquisition`, `isnad-graph`, `user-service`) have the identical `apt-get -y upgrade` + `cache-from: type=gha` + no `no-cache-filter` shape and are one advisory away from the same failure. Their green is not evidence of coverage.

## Secondary finding — the runtime image ships the build toolchain

Separate from the cache defect and worth fixing in the same pass. The top-level `Dockerfile` is **single-stage** and installs `build-essential curl` into the image that ships:

- `build-essential` → `libc6-dev` → **`linux-libc-dev`**, the source of 7 of the 10 CVEs (kernel headers, in a container that will never compile a kernel).
- `curl` → **`libssh2-1t64`**, the source of the other 3.

[...]

PR #129, opened 2026-07-11
fix(ci): clear pip-audit advisory-DB drift — bump soupsieve, exempt unfixable ecdsa Minerva advisory

### LINKED ISSUE #128: ci(security-audit): pip-audit red on advisory-DB drift — soupsieve CVEs (fixable) + unfixable ecdsa Minerva advisory blocks every PR
## Summary

`security-audit` (`pip-audit --strict`) now fails on `main` against an **unchanged `uv.lock`**. Three advisories published after 2026-07-06 and reddened a dependency set nobody touched. This blocks **every** PR in this repo — including a no-op one — and it is currently blocking #127.

This is advisory-DB drift, not a code regression.

## Evidence

`ci.yml` on `main` was **success** at `fd698ca` (2026-07-06) with the identical lockfile. The same export today reports:

| Package | Version | Advisory | Fix |
|---|---|---|---|
| `soupsieve` | 2.8.3 | CVE-2026-49476 (unbounded alloc on comma-separated selector lists) | **2.8.4** |
| `soupsieve` | 2.8.3 | CVE-2026-49477 (ReDoS on unterminated quoted attribute value) | **2.8.4** |
| `ecdsa` | 0.19.2 | PYSEC-2026-1325 (Minerva timing attack, P-256) | **none — upstream declines** |

`soupsieve` arrives transitively via `beautifulsoup4`. `ecdsa` arrives via `python-jose`, which is a real direct dependency (`src/api/auth.py` imports `jwt` from it).

## Fix

1. **`soupsieve` → 2.8.4.** A real fix for two real CVEs. No judgment call.

2. **`--ignore-vuln PYSEC-2026-1325` for `ecdsa`.** The advisory has no fix and upstream has stated it will not produce one (python-ecdsa considers side-channel attacks out of scope). `ecdsa` cannot be dropped without dropping `python-jose`, which the admin API genuinely uses.

   **Why the exemption is safe — the vulnerable primitive is unreachable.** `src/api/auth.py:123,129` calls `jwt.decode(token, jwks, algorithms=["RS256"])`. The algorithm allowlist is **RS256 — RSA, not ECDSA**. The advisory is a timing attack on ECDSA P-256 scalar multiplication, a code path this repo never reaches in either direction. `ecdsa` is installed only because `python-jose` declares it, not because we call it.

   **What would revoke this exemption:** changing that algorithm allowlist to admit an EC algorithm (`ES256`/`ES384`/`ES512`). If anyone does that, the ignore must be reconsidered — the "unreachable primitive" rationale dies at that moment.

   Precedent: the same command line already carries `--ignore-vuln GHSA-g75f-g53v-794x` for the unfixable bleach advisory, with a rationale comment and a standing item (noorinalabs-main#703).

Both `ci.yml` (`security-audit`) and `.pre-commit-config.yaml` (the pre-push `pip-audit` hook) must be changed together — they mirror each other verbatim, and local/CI tooling parity is mandatory (noorinalabs-main#684).

## Notes verified while diagnosing

- **`PYSEC-2026-1325` is an alias of `CVE-2024-23342`** (OSV `aliases: [CVE-2024-23342, GHSA-wj6h-64fc-37mp]`) — the original python-ecdsa Minerva advisory, not a new one.
- **`pip-audit --ignore-vuln` matches aliases.** Verified empirically: ignoring by `CVE-2024-23342` suppresses the `PYSEC-2026-1325` finding identically to ignoring by the reported id.
- **`noorinalabs-isnad-graph` is therefore already covered and needs no change** — its existing `--ignore-vuln CVE-2024-23342` suppresses this advisory by alias, and it carries no `beautifulsoup4`/`soupsieve`.
- `noorinalabs-data-acquisition` carries `soupsieve` 2.8.3 but has **no** `pip-audit` invocation in its workflows, so it cannot red this way.

## Acceptance

- [ ] `pip-audit --strict` exits 0 locally and in CI against the updated lock.
- [ ] The `ecdsa` ignore carries an inline rationale naming the RS256-only allowlist and the revocation condition — a bare CVE id with no justification is what made this expensive to diagnose.
- [ ] `ci.yml` and `.pre-commit-config.yaml` stay verbatim-identical.
- [ ] A standing item records the exemption so it is revisited rather than forgotten.

parametrization:
## What

Clears the `security-audit` (`pip-audit --strict`) red that is currently **blocking every PR in this repo**, including #127.

This is **advisory-DB drift, not a code regression**: `ci.yml` was green on `main` at `fd698ca` (2026-07-06) with this **exact same `uv.lock`**. Three advisories published afterwards against a dependency set nobody touched.

## Changes

**1. `soupsieve` 2.8.3 → 2.8.4** (lockfile only)

Fixes CVE-2026-49476 (unbounded allocation on comma-separated selector lists) and CVE-2026-49477 (ReDoS on an unterminated quoted attribute value). Transitive via `beautifulsoup4`. A real fix for real CVEs — no judgment call.

**2. `--ignore-vuln PYSEC-2026-1325` for `ecdsa`** — with an inline rationale

No fix exists and **none is coming**: the python-ecdsa project considers side-channel attacks out of scope. `ecdsa` cannot be dropped — `python-jose` pulls it, and `src/api/auth.py` genuinely uses `python-jose`.

The exemption is safe because **the vulnerable primitive is unreachable**, not merely unused in one direction. `src/api/auth.py:123,129` calls `jwt.decode(token, jwks, algorithms=["RS256"])` — **RS256 is RSA**. The advisory is a timing attack on **ECDSA P-256 scalar multiplication**, a code path this repo never enters. `ecdsa` is installed only because `python-jose` declares it as a dependency.

> Deliberately **not** justified as "we only verify, we don't sign." That is true but brittle — it would survive a switch to ES256 while the actual risk would not.

**What revokes this exemption:** admitting any EC algorithm (`ES256`/`ES384`/`ES512`) into that allowlist. That condition is written into the inline comment **and** into a standing item (`.claude/memory/project_ecdsa_minerva_standing_item.md`), so the next person knows what would invalidate it — rather than finding a bare CVE id with no justification, which is exactly what made this expensive to diagnose.

Added to **both** `ci.yml` and `.pre-commit-config.yaml`; they mirror each other verbatim (mandatory local⇄CI tooling parity, noorinalabs-main#684).

## Verified, not assumed

- **`PYSEC-2026-1325` is an alias of `CVE-2024-23342`** (OSV `aliases: [CVE-2024-23342, GHSA-wj6h-64fc-37mp]`) — the original python-ecdsa Minerva advisory, not a new one.
- **`pip-audit --ignore-vuln` matches aliases.** Confirmed empirically: ignoring by `CVE-2024-23342` suppresses the finding reported as `PYSEC-2026-1325`, identically to ignoring by the reported id.
- **`noorinalabs-isnad-graph` therefore needs no change** — its pre-existing `--ignore-vuln CVE-2024-23342` already suppresses this advisory by alias, and it carries no `beautifulsoup4`/`soupsieve`. No issue filed there; there is nothing to fix.
- **`noorinalabs-data-acquisition`** carries `soupsieve` 2.8.3 but has **no** `pip-audit` invocation in its workflows, so it cannot red this way.

## Test plan

- `pip-audit --strict` **exits 0** locally against the updated lock (it exited 1 before). The push itself is evidence: `pip-audit` is a pre-push hook, so this branch could not have been pushed while red.
- Full local check-set green, both stages: ruff-format, ruff-lint, actionlint, cspell, dockerfile-base-pin, structural-ontology, sync-drift gate, mypy --strict, pytest.
- No source change — `pytest` and `mypy` cover that nothing moved.

Closes #128
Unblocks #127

### DIFF
--- a/.claude/memory/MEMORY.md
+++ b/.claude/memory/MEMORY.md
@@ -4,3 +4,4 @@ This is the always-loaded index (one line per memory). Topic files in this direc
 
 - [Data pipeline arch](project_data_pipeline_architecture.md) — B2 storage, Kafka/KRaft pipeline, acquisition-vs-platform repo split, MinIO local dev.
 - [main#136 E2E lands in ingest](project_main136_pipeline_e2e_lands_in_ingest.md) — main#136 pipeline-worker E2E scenarios land in ingest-platform (not deploy/not main), ingest persona.
+- [ecdsa Minerva standing item](project_ecdsa_minerva_standing_item.md) — PYSEC-2026-1325 pip-audit-ignored: RS256-only allowlist makes the ECDSA path unreachable; switching to ES* REVOKES it. pip-audit --ignore-vuln matches aliases. ingest#128.
--- a/.claude/memory/project_ecdsa_minerva_standing_item.md
+++ b/.claude/memory/project_ecdsa_minerva_standing_item.md
@@ -1,0 +1,22 @@
+---
+name: project_ecdsa_minerva_standing_item
+description: ecdsa Minerva advisory (PYSEC-2026-1325) is pip-audit-ignored because RS256-only auth makes the ECDSA path unreachable; switching to ES* revokes the exemption.
+metadata:
+  type: project
+---
+
+`pip-audit --strict` ignores **PYSEC-2026-1325** (aliases **CVE-2024-23342**, **GHSA-wj6h-64fc-37mp**) — the python-ecdsa **Minerva timing attack on P-256**. Ignored in BOTH `ci.yml` (`security-audit`) and `.pre-commit-config.yaml` (pre-push `pip-audit`); they must stay verbatim-identical (local⇄CI parity, noorinalabs-main#684). Filed as ingest#128, 2026-07-10.
+
+**There is no fix and none is coming** — the python-ecdsa project considers side-channel attacks out of scope. `ecdsa` cannot be dropped: `python-jose` pulls it, and `src/api/auth.py` genuinely uses `python-jose`.
+
+**Why the exemption is sound:** the vulnerable primitive is **unreachable**, not merely unused in one direction. `src/api/auth.py:123,129` calls `jwt.decode(token, jwks, algorithms=["RS256"])` — RS256 is **RSA**. The advisory is a timing attack on **ECDSA P-256 scalar multiplication**, a code path this repo never enters. `ecdsa` is installed only because `python-jose` declares it as a dependency, not because anything calls it.
+
+> Do not restate this as "we only verify, we don't sign." That weaker rationale is true but brittle — it would survive a switch to ES256 while the actual risk would not.
+
+**What REVOKES this exemption:** admitting any EC algorithm (`ES256` / `ES384` / `ES512`) into that `algorithms=[...]` allowlist. The moment auth can verify an EC-signed token, the "unreachable primitive" argument dies and the ignore must be re-justified or removed. Revisit each wave.
+
+**Two mechanics worth remembering (both verified, not assumed):**
+- **`pip-audit --ignore-vuln` matches ALIASES.** Ignoring by `CVE-2024-23342` suppresses a finding reported as `PYSEC-2026-1325`, and vice versa. So an ignore list keyed on any one alias covers the others.
+- Consequently **`noorinalabs-isnad-graph` was already covered** — its pre-existing `--ignore-vuln CVE-2024-23342` suppresses this advisory by alias, and it carries no `beautifulsoup4`/`soupsieve`. No change was needed there.
+
+Same shape as the bleach standing item (`GHSA-g75f-g53v-794x`, noorinalabs-main#703): an unfixable upstream advisory, non-applicable to our usage, ignored with an inline rationale and a revocation condition. Related: [[project_data_pipeline_architecture]].
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,25 @@ jobs:
           # upstream fix released. parse_email is not used anywhere in this repo
           # (grep -rn parse_email is empty) → non-applicable. Tracked in
           # noorinalabs-main#703; revisit each wave for an upstream fix.
-          uv run pip-audit --strict --desc --ignore-vuln GHSA-g75f-g53v-794x -r /tmp/requirements.txt
+          #
+          # PYSEC-2026-1325 (aliases CVE-2024-23342 / GHSA-wj6h-64fc-37mp):
+          # python-ecdsa Minerva timing attack on P-256. No upstream fix and none
+          # coming — the project considers side-channel attacks out of scope.
+          # `ecdsa` cannot be dropped: it is pulled by python-jose, which the
+          # admin API genuinely uses.
+          #
+          # Non-applicable because the vulnerable primitive is UNREACHABLE, not
+          # merely unused in one direction: src/api/auth.py:123,129 calls
+          # jwt.decode(token, jwks, algorithms=["RS256"]). The allowlist is
+          # RS256 — RSA. The advisory is a timing attack on ECDSA P-256 scalar
+          # multiplication, a code path this repo never enters. `ecdsa` is
+          # installed only because python-jose declares it.
+          #
+          # REVOKES THIS EXEMPTION: admitting an EC algorithm (ES256/ES384/
+          # ES512) into that allowlist. If you change it, this ignore must be
+          # re-justified — the "unreachable primitive" rationale dies there.
+          # Tracked in ingest#128; revisit each wave.
+          uv run pip-audit --strict --desc --ignore-vuln GHSA-g75f-g53v-794x --ignore-vuln PYSEC-2026-1325 -r /tmp/requirements.txt
 
   test:
     runs-on: ubuntu-latest
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -149,7 +149,17 @@ repos:
         # ReDoS — no upstream fix; parse_email unused in this repo → non-
         # applicable. Tracked in noorinalabs-main#703; revisit each wave. Kept
         # identical to ci.yml's security-audit invocation (local⇄CI parity).
-        entry: bash -c 'uv export --no-hashes --frozen --no-emit-project > /tmp/requirements.txt && uv run pip-audit --strict --desc --ignore-vuln GHSA-g75f-g53v-794x -r /tmp/requirements.txt'
+        #
+        # --ignore-vuln PYSEC-2026-1325 (aliases CVE-2024-23342 /
+        # GHSA-wj6h-64fc-37mp): python-ecdsa Minerva timing attack on P-256. No
+        # upstream fix and none coming (side-channels are out of scope for that
+        # project). Non-applicable because the vulnerable primitive is
+        # unreachable: src/api/auth.py:123,129 decodes with
+        # algorithms=["RS256"] — RSA, not ECDSA — so the P-256 scalar-mult path
+        # is never entered. `ecdsa` is present only because python-jose declares
+        # it. Admitting an EC algorithm (ES256/ES384/ES512) into that allowlist
+        # REVOKES this exemption. Tracked in ingest#128; revisit each wave.
+        entry: bash -c 'uv export --no-hashes --frozen --no-emit-project > /tmp/requirements.txt && uv run pip-audit --strict --desc --ignore-vuln GHSA-g75f-g53v-794x --ignore-vuln PYSEC-2026-1325 -r /tmp/requirements.txt'
         language: system
         pass_filenames: false
         always_run: true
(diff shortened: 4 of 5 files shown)


parametrization:
## Security review — Kalinda Ranasinghe

I re-verified both load-bearing claims at the source rather than accepting them. Both hold, and the unreachability argument is in fact **stronger** than the one written in the comment. Approving.

### 1. `soupsieve` 2.8.3 → 2.8.4

Not a judgment call — two HIGH CVEs, upstream fix exists, take the fix. No notes.

### 2. `--ignore-vuln PYSEC-2026-1325` (ecdsa 0.19.2, Minerva)

Suppressing a finding is a policy act, so I checked it as one. Four independent barriers, each sufficient on its own:

**(a) `RS256` is genuinely the only algorithm admitted, on every path.** Grepped the whole tree at head `ba8e843` (`jwt.decode|jwt.encode|jws.|algorithms=|get_unverified|from jose`), not just the two lines cited:

```
src/api/auth.py:24    from jose import JWTError, jwt
src/api/auth.py:123       jwt.decode(token, jwks, algorithms=["RS256"])
src/api/auth.py:129       jwt.decode(token, jwks, algorithms=["RS256"])
```

That is the complete set. Two call sites, both explicit allowlists, no third site, no default-permits-more, no `get_unverified_header` steering. python-jose rejects a token whose header `alg` is outside the allowlist *before* it constructs a key, so an attacker-supplied `ES256` token dies at the allowlist and never reaches EC code.

**(b) The repo never signs.** `jwt.encode`, `SigningKey`, `sign_digest`, `private_key` → **zero** matches across `src/`, `workers/`, `tests/`. user-service is the sole issuer, as intended. This matters more than the PR gives it credit for — see the correction below.

**(c) `ecdsa` has exactly one dependent.** One `{ name = "ecdsa" }` edge in `uv.lock`, under `python-jose` 3.5.0. Nothing else pulls it.

**(d) The one I want on the record — `ecdsa` is not even the backend.** The repo declares `python-jose[cryptography]` plus an explicit `cryptography>=48.0.1`, so jose resolves EC to the cryptography backend, and the `ecdsa` package is never imported at runtime:

```
ECKey class -> jose.backends.cryptography_backend.CryptographyECKey
ecdsa module imported at runtime? False
```

So even if someone *did* admit `ES256` tomorrow, the EC verify would route through `cryptography`, not through the vulnerable package. `ecdsa` is installed-but-dead weight.

### 3. The alias claim — re-verified, it holds

This was the part that could have left a silent gap in isnad-graph, so I tested the instrument on both classes instead of reading a clean result and calling it proof.

OSV confirms the linkage (`PYSEC-2026-1325` → aliases `CVE-2024-23342`, `GHSA-wj6h-64fc-37mp`, affected package `ecdsa`). Then, against a pinned `ecdsa==0.19.2`:

| invocation | result |
|---|---|
| no ignore (control — detector must fire) | `Found 1 known vulnerability` |
| `--ignore-vuln PYSEC-2026-1325` (this PR) | `No known vulnerabilities found, 1 ignored` |
| `--ignore-vuln CVE-2024-23342` (what isnad-graph already carries) | `No known vulnerabilities found, 1 ignored` |

The control fires, so the two clean runs are measurements and not vacuous. `pip-audit --ignore-vuln` **does** match aliases. isnad-graph's pre-existing ignore genuinely covers this advisory; filing nothing against it was correct, and there is no silent gap waiting to red on its next run.

Local⇄CI parity intact — the `ci.yml` and `.pre-commit-config.yaml` invocations are verbatim-identical (noorinalabs-main#684).

### One correction, non-blocking

The memory file and the inline comments both say the "we only verify, we don't sign" framing is *brittle* because it "would survive a switch to ES256 while the actual risk would not." The advisory text says the opposite:

> Both ECDSA signatures, key generation, and ECDH operations are affected. **ECDSA signature verification is unaffected.**

The vulnerable operation is *signing* (nonce leak via `sign_digest`). Verification — including ES256 verification — is not affected at all. So barrier (b) is not the weak one; it is arguably the strongest, and admitting `ES256` would not by itself materialize the risk.

This does not change the verdict and I am **not** asking you to loosen the revocation condition. A revocation trigger that fires earlier than strictly necessary costs one re-review and never leaves an exemption silently outliving its rationale — that is the correct direction to err. But the stated *reason* misdescribes the advisory, and a future engineer will read that memory and carry a wrong model of what Minerva actually breaks. Worth a small in-place edit to the memory file when convenient: keep the same trigger, fix the justification.

### Verdict

The exemption is sound and correctly framed as consistent application of an existing decision, not a new weaker one. The revocation condition is recorded, in both files, at the call site — which is more than the isnad-graph ignore it is modeled on ever had. That is the standard I want every future suppression held to.

Approved.

---
Requestor: Kalinda Ranasinghe
Requestee: Bjørn Henriksen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Review — correctness / dependency-hygiene axis

Approved. This is a correct, minimal, well-evidenced fix and I don't want to hold it up — ip#127 is queued behind it. Everything below is verification, not objection, except one finding that is about our tooling rather than about this PR.

### The lock diff is clean

`uv.lock` moves `soupsieve` and nothing else. The whole diff is a single `[[package]]` block — version, sdist hash, wheel hash — and I confirmed no second package block is touched. An `--upgrade-package` can quietly drag siblings; this one did not.

That matters more here than usual because the `Dockerfile` installs with `--require-hashes` (lines 43–44). It regenerates the hashed requirements from `uv.lock` at build time via `uv export --frozen --format requirements-txt`, so the lock *is* the hash source for the image. A silently-changed hash on some other package would have flowed straight into the build. None did — the only hashes entering the image that changed are `soupsieve`'s, and they match the lock.

### 2.8.4 is genuinely the fixed version, and it is fixed rather than suppressed

`beautifulsoup4` 4.14.3 requires `soupsieve` with **no version bound at all** (`{ name = "soupsieve" }`), so 2.8.4 is admitted trivially. `beautifulsoup4` is its only dependent, so nothing else can pin it back.

The stronger evidence is structural. `pip-audit --strict` is green at head with an ignore list containing exactly two entries — the bleach advisory and the ecdsa one. **Neither is a `soupsieve` advisory.** So a green audit at 2.8.4 is `pip-audit` reading the live advisory DB and reporting zero findings against that version. The CVEs were cleared by the bump; they could not have been cleared by suppression, because the suppression list doesn't contain them. That is a better proof than any changelog would be, and it's worth noting it comes for free from the way you structured the change.

### The ignore flag is mirrored exactly

I compared the two invocations rather than the two flag-lists, because the export step matters too. Both sides run `uv export --no-hashes --frozen --no-emit-project` and then `pip-audit --strict --desc --ignore-vuln GHSA-g75f-g53v-794x --ignore-vuln PYSEC-2026-1325 -r /tmp/requirements.txt`. Same export flags, same audit flags, same order. Parity holds byte-for-byte between `ci.yml` and `.pre-commit-config.yaml`.

### The sync gate does **not** prove that, and its green is not evidence

Worth stating plainly, because it's easy to assume otherwise. `precommit-ci-sync` is green here, but that green would also have appeared if you had put the new ignore in only one of the two files.

`pre_commit_ci_sync.py` classifies each side into a set of check *kinds* and does set arithmetic (`harmful = ci_kinds - precommit_kinds`). The **arguments are never compared**. `pip-audit` is present on both sides, so the kind matches, so the gate passes — whatever the flags say. Your parity is real, but it was established by reading both files, not by the gate.

This is a different blind spot from the unclassified-kind one that main#684 tracks: this is a *classified* kind with unverified arguments. The gate sees it, reports on it, and is still blind. I filed it as noorinalabs-main#949 rather than asking you to fix shared org tooling from a two-line deps PR. Not a blocker on this PR.

### The standing item will not rot

It names a revocation condition and a revisit trigger, which was my main worry — an `--ignore-vuln` with neither becomes a permanent unexplained suppression, and six months from now nobody can tell whether it's still justified or just inherited.

I also verified the load-bearing claim instead of taking the memory file's word for it. `src/api/auth.py:123,129` really does call `jwt.decode(token, jwks, algorithms=["RS256"])`, and that file is the **only** one of the repo's 183 Python files containing an `algorithms=` allowlist at all. There is no `ES256`/`ES384`/`ES512` anywhere in the repo. So the "unreachable primitive" rationale is sound as written, and the cited line numbers are accurate.

I want to call out the reasoning in that note specifically. Declining to justify the exemption as "we only verify, we don't sign" — because that rationale is true but would survive a switch to ES256 while the actual safety would not — is the right call, and it's the difference between a revocation condition that tracks the real risk and one that merely looks rigorous. The condition you did write (admitting an EC algorithm into that allowlist) is exactly the event that kills the argument.

### Nothing blocking

One observation, no action needed: the audit path exports without `--no-dev`, while the image build exports with it, so `pip-audit` audits a superset of what ships. That's the conservative direction and I'd leave it alone.

Good evidence discipline throughout — proving the advisories were pre-existing by pinning CI green at the identical lock on the base SHA, and testing the alias-matching behaviour empirically rather than trusting the docs, are both what I'd want and neither is the default. Ship it.

---
Requestor: Petra Vidović
Requestee: Bjørn Henriksen
RequestOrReplied: Approved
TechDebt: noorinalabs-main#949

PR #130, opened 2026-07-11
fix(ci): force the apt layer to rebuild every publish (no-cache-filter: base) — cache fix ALONE

### LINKED ISSUE #127: ci(publish): GHCR publish red on main since 07-06 — cached apt layer freezes 10 HIGH OS CVEs (repo-side of main#947)

parametrization:
> **Rewritten after review.** Tomás Carvalho blocked the original version of this PR, correctly: it bundled the toolchain removal with the cache fix, and in doing so **destroyed its own proof**. This PR is now the cache fix *alone*. The toolchain removal follows in a separate PR. His reasoning is reproduced below because it is the whole design of this change.

## What

`Publish to GHCR` has been **red on `main` since 2026-07-06** — five days. Trivy fails on 10 HIGH OS-package CVEs, **all of which have published fixes**.

The fixes are never installed. The buildx GHA cache serves the `RUN apt-get update && apt-get -y upgrade` layer from cache, so the upgrade is replayed as a **no-op**:

- **`#11 CACHED`** on the apt layer (job `86502590623`), imported from `gha:2684583647406898998`.
- Package versions **byte-identical** across builds five days apart: `libssh2-1t64` still `1.11.1-1` (fixed in `1.11.1-1+deb13u1`), `linux-libc-dev` still `6.12.94-1` (fixed in `6.12.95-1`).
- HIGH count **grew 5 → 10** against the frozen layer. A build that actually re-ran apt would have gone *down*.

The charter's declared defense against **within-tag package drift** (`tech-decisions.md` § Base Image Pinning) is structurally inert, and `check_dockerfile_base_pin.py` cannot see it — that gate proves the upgrade **line exists**, not that it **runs**.

### This is a class defect, not a quirk of this repo

`noorinalabs-data-acquisition`'s publish run [29138030830](https://github.com/noorinalabs/noorinalabs-data-acquisition/actions/runs/29138030830) — a **green** run, different cache scope (`load`), already multi-stage, `gcc libpq-dev` rather than `build-essential` — logs:

```
#11 [base 2/2] RUN apt-get update     && apt-get -y upgrade     && apt-get install -y --no-install-recommends gcc libpq-dev     && rm -rf /var/lib/apt/lists/*
#11 CACHED
```

ingest-platform is not special. It is the repo that has been *hit*; da's apt layer is equally frozen, its green means only that nothing has shot at it yet. Measured on **2 of the 4** affected repos.

## The change

Name the base stage (`AS base`) in the five Dockerfiles, and pass `no-cache-filter: base` to `docker/build-push-action`. **That is all.**

## Why the diff is deliberately minimal — the diff *is* the experiment

`AS` names a stage. It does **not** change the resolved image digest or any instruction's text. The `ENV` and `RUN` in the base stage are left **byte-identical to `main`** — verified: the apt `RUN` block hashes the same before and after, and no `ENV`/`RUN`/`COPY` line appears in the diff.

So **every cache key in the base stage is unchanged**, and the cached layer from run `28766990515` remains a **valid hit**.

That is the entire point. The only thing that can now make the apt layer rebuild is **`no-cache-filter` actually working**. The first post-merge publish becomes a clean, immediately-discriminating experiment with **two independent signals that must agree**:

1. **The apt layer is not `CACHED`** — meaningful *precisely because* the layer's content did not change and it therefore **would** have hit.
2. **Trivy goes green** — independently meaningful because **`build-essential` and `curl` are still installed in this PR**, so `linux-libc-dev` and `libssh2-1t64` are still in the image and still scanned. All 10 CVEs are `fixed` upstream and the scan sets `ignore-unfixed: true`, so **they clear if and only if apt actually upgraded them**.

And the failure mode is **loud**: if the flag is inert, `main` stays **red** and the system tells us immediately, while the evidence still exists.

## Why the toolchain removal is NOT in this PR

The original PR bundled it. That would have:

- **Confounded the `CACHED` signal** — dropping `UV_SYSTEM_PYTHON`/`UV_LINK_MODE` from the `ENV` above changes the parent vertex, and dropping `build-essential curl` from the `RUN` changes the instruction text. Either alone is a cache miss on **content** grounds, so the layer rebuilds *regardless* of `no-cache-filter` and the log line means nothing.
- **Confounded the Trivy signal** — it deletes the very packages all 10 CVEs live in, so Trivy could go green even if the flag did nothing.

**A silently-inert flag and a working one would have produced byte-identical evidence.** And had it shipped, `main` would have gone green — the five-day-red gate cleared, a permanent tick on the commit — with the only contradicting artifact being one build-log line that expires in 90 days and that a human must volunteer to go read against his own passing PR. A fix whose correctness rests on someone choosing to contradict a green check is exactly the design this issue exists to retire. ingest is the **pathfinder** for the da / isnad-graph / user-service rollout; getting it wrong here propagates.

The toolchain removal is good hardening and it *is* coming — it just must not eat the evidence for the fix it ships with.

## PROVEN — pre-merge, with nothing published

The previous revision of this branch was **inert**. It passed `no-cache-filter` (**singular**), which is *not a declared input* of `docker/build-push-action` — it is the docker CLI flag the action *emits*. Verified at the exact pinned SHA `f9f3042`:

```
action.yml            declares only `no-cache` and `no-cache-filters`
src/context.ts:72     getInputList('no-cache-filters')            <- reads PLURAL
src/context.ts:163-4  loops that list -> args.push('--no-cache-filter', ...)
```

With the singular key that list is **empty**, the loop runs **zero** times, buildx never sees the flag. Nothing catches it: **GitHub Actions does not fail on an undeclared `with:` key** (warning annotation, silently dropped), and **actionlint does not validate third-party action input schemas**. The inert form lints green, builds green, and leaves the apt layer frozen. Caught by **Fatima Bensalah**, reading the pinned action's source.

That failure mode is worse than a loud one: it fails *safe* (main stays red), so nothing shouts — but the natural reading becomes *"we applied the filter and the CVEs are still there, so the cache wasn't the problem."* **An inert fix discredits a correct diagnosis.** This repo is the pathfinder for da / isnad-graph / user-service, so it would have shipped four times.

### The evidence — run [29139568901](https://github.com/noorinalabs/noorinalabs-isnad-ingest-platform/actions/runs/29139568901)

A temporary harness built this **exact Dockerfile** against the **same `scope=workers` GHA cache**, `push: false` / `load: true` / **no `cache-to`** — so nothing reached GHCR and **`stg-latest` never moved**. It ran the build **twice**, control vs treatment, because "the layer rebuilt" is only evidence if it *would otherwise have been cached*:

> **Corrected after review (Fatima Bensalah).** An earlier version of this table compared **buildx vertex numbers**. Those are assigned in *solve order* and are **not stable across builds** — control's `#10` is `[base 5/9] COPY pyproject.toml uv.lock`, while treatment's `#10` is the apt layer. The comparison was invalid as constructed and reached the right answer **by luck**. Re-derived below against the stable `[base N/9]` **stage-step label**. An unstable identifier can fabricate a refutation as easily as a confirmation.

| stage-step | control (no filter) | treatment (`no-cache-filters: base`) |
|---|---|---|
| `[base 2/9] RUN apt-get … upgrade` | **CACHED** | **DONE 11.8s** — apt executed |
| `[base 6/9] RUN uv … pip install` | CACHED | DONE 3.5s |
| Trivy, `debian 13.5` target | **10** | **0** |

The conclusion is unchanged and now rests on a stable identifier: **the apt layer was a valid cache hit and the filter forced it to rebuild.**

The treatment log shows it fetching the exact fixed packages:

```
#10 Get:17 .../trixie-security/main amd64 linux-libc-dev all 6.12.95-1 [2852 kB]
#10 Get:72 .../trixie-security/main amd64 libssh2-1t64 amd64 1.11.1-1+deb13u1 [245 kB]
```

Both signals move, in both directions, and they are **independent** — `build-essential`/`curl` are still installed in this PR, so `linux-libc-dev` and `libssh2-1t64` are still in the image and still scanned. They cleared because apt actually upgraded them. The harness is **deleted** from the branch; the run logs are the artifact.

### Known, temporary: the gha cache is now dead config

Because the Dockerfile is still **single-stage**, `base` *is* the whole image — so excluding it uncaches **everything**. Expect a publish log with **no `CACHED` lines at all**; that is correct, not anomalous.

**The cost is ~15s of layer work** (job wall-clock delta ~12s). An earlier revision of this PR — and of the committed comment — claimed **~80s**. That was **wrong**: the 43.4s it rested on is `exporting to docker image format`, a step that exists **only** because the proof harness used `load: true`. A real `push: true` publish never runs it. Caught by Fatima.

### The `cache-from`/`cache-to` lines STAY — they are the detector, not an optimisation

I initially argued to keep them for speed, then (once the uv install turned out to be 3.5s) argued to **delete** them. **Both were wrong, and Tomás corrected me.** Their value has nothing to do with speed:

- **cache import KEPT + the filter later breaks** (stage renamed, the action renames its input, someone "tidies" the key back to the singular): buildx serves the frozen apt layer → stale packages → **Trivy goes RED** → we find out that day.
- **cache import DELETED + the filter breaks the same way**: nothing to serve, apt re-runs anyway, **Trivy goes GREEN**, and the dead flag is **invisible** — rotting until #132's split makes the filter load-bearing.

That second row is **this exact bug in a new costume: a guard that goes quiet in precisely the case it exists to catch.**

And the sharpest part: **today the security property does not rest on the filter at all** — `base` is the whole image, so nothing could serve a stale layer even if the filter were dead. The filter only starts carrying weight *after* #132's split. **The cache import is how we will know the filter still works on the day it starts to matter.**

**`cache-to` is what keeps the detector alive** (Tomás): GHA cache entries evict when unused, so dropping **only** `cache-to` silently empties `cache-from` and kills the guard **without anyone touching it**. That is the same defect again — a detector that decays into silence.

### And the decision must not be priced at all

**Every number attached to this so far was measured in the wrong arm**, mine included:

- my **~80s** — a `load: true` harness artefact (`exporting to docker image format`), which a real `push: true` publish never runs;
- my **3.5s** uv saving — exists **only in the control arm**, and is unreachable once the filter is on;
- and a proposed **"zero"** — refuted by Tomás against his own conclusion, because `cache-to: mode=max` still *runs* every publish and, since the filter rebuilds every layer, re-exports the **full** layer set each time.

**The real cost is the `cache-to` export, and it is UNMEASURED — the proof harness omitted `cache-to` by design.** The one arm that would have priced it is the one arm nobody ran. So the comment prices nothing: the argument for keeping these lines is structural and never depended on a number. I'll read the export cost off the first real publish.

Whether the split *also* restores caching as a speed win is **an open question, and it stays open in both directions** — it may cascade (`builder` derives `FROM base`), or it may not (BuildKit chains a child's key on the parent's *result*, so a byte-identical re-run still lets children hit, and a genuine upgrade correctly busts them). **Nobody's belief goes in the comment.** It gets measured in #132, and either answer is a fact about *speed* that licenses deleting nothing.

## Test plan

Local check-set green, both stages (ruff, actionlint, cspell, dockerfile-base-pin, structural-ontology, sync-drift, mypy --strict, pytest, pip-audit).

Its green checks are **not** what establishes correctness here — the proof above is (run `29139568901`), and it was obtained **without publishing anything**.

After merge I will still confirm on the real `main` publish log that the apt vertex is not `CACHED` and Trivy reports 0 HIGH/CRITICAL on the pushed digest — but that is now a confirmation of a measured result, not the first test of an untested one.

Closes #127
Part of noorinalabs-main#947


parametrization:
## Review — falsifiability / evidence axis

Fatima has the Docker/CI mechanics. I took the evidence axis: *does the proof this PR offers actually discriminate between "the fix works" and "the fix does nothing"?*

The code is right. I could not find a defect in it. **The verification plan, however, does not work** — and it fails in the exact way this PR exists to eliminate. That is the one blocking item.

Bjørn deserves credit up front: he identified the Trivy confound himself and wrote it into the body rather than banking a green tick. That is the right instinct and it is why this review is about a second confound rather than the first.

---

## Blocking: the stated proof cannot fail on the build it will be read from

The PR body names one discriminating observation:

> **Only the absence of `CACHED` on the apt layer proves the actual defect is fixed.**

That is the correct *kind* of observation. But on the first post-merge `main` publish — the build Bjørn plans to read it from — **it is guaranteed to hold whether or not `no-cache-filter: base` does anything at all.**

BuildKit caches per-instruction, not per-file. That is exactly what run `28766990515` showed: `#11 CACHED`, one vertex. The cache key for the apt `RUN` is a hash of its parent layer plus its own instruction text. This PR changes **both** of those inputs in the `base` stage:

- the `ENV` immediately above it drops `UV_SYSTEM_PYTHON=1` and `UV_LINK_MODE=copy` — the parent vertex changes;
- the `RUN` itself drops `&& apt-get install -y --no-install-recommends build-essential curl` — the instruction text changes.

Either alone is a cache miss. So the apt layer **must** rebuild on the first publish, on content grounds, with `no-cache-filter` playing no part. It will come back not-`CACHED`, the log line will read exactly as predicted, and it will mean nothing.

So as written, on that first publish:

- **Trivy is confounded** — the packages the 10 CVEs live in are deleted (Bjørn's own finding).
- **The `CACHED` line is confounded** — the layer's content changed, so it rebuilds regardless.

Both signals are destroyed, by two independent mechanisms. The PR would merge with **no observation available that could come out differently if `no-cache-filter: base` were a typo.** A silently-inert flag and a working one produce byte-identical evidence. That is not a weak proof; it is not a proof.

This is the same species as the gate that certifies the `apt-get upgrade` *line exists* while the cache stops it *running* — reproduced one level up, inside the fix's own acceptance criteria.

`no-cache-filter` only demonstrates anything on a build where the `base` stage content is **unchanged** from a cached predecessor. That is the only condition under which "would have been `CACHED`, wasn't" is a fact rather than a tautology.

## What would falsify the fix

Stating this plainly, since the brief asked for it, and because the PR should not merge without it written down somewhere a future reader will find:

> **The apt layer in the `base` stage appearing as `CACHED` in a publish whose `base` stage content is byte-identical to the preceding publish's.**

The qualifier is load-bearing. Without it the observation is unfalsifiable on exactly the run it will be checked against.

## Requested change — split the PR, cache fix first

Yes, this should be two PRs. Not on hygiene grounds — on the grounds that **the split is what makes the experiment work**, and it costs one publish cycle.

**PR-A — the cache fix, alone.** `AS base` on the five `FROM` lines, plus `no-cache-filter: base` in `ghcr-publish.yml`. Nothing else. No `ENV` change, no `RUN` change, no multi-stage split.

The `AS` clause names a stage; it does not change the resolved image digest or any instruction's text, so **every cache key in the `base` stage stays identical**. The cached apt layer from run `28766990515` is still a valid hit. Which means the only thing on earth that can make that layer rebuild is `no-cache-filter` actually working — and the first publish becomes a clean, immediately-discriminating experiment:

1. **Apt layer not `CACHED`** — now meaningful, because the content was unchanged and it *would* have hit.
2. **Trivy goes green** — independently meaningful, because `build-essential` and `curl` are still installed, so `linux-libc-dev` and `libssh2-1t64` are still in the image and still scanned. All 10 are `fixed` upstream and the scan runs `ignore-unfixed: true`, so they clear **if and only if apt actually upgraded them.**

Two independent signals that agree, on the first build, from a fix that is about to be replicated to three more repos. If instead the publish stays red, we learn that immediately and while the evidence still exists.

And the failure mode becomes **loud**. That is the real argument. If `no-cache-filter` is inert and PR-A ships alone, `main` stays red and the system tells us. If it is inert and this PR ships as-is, `main` goes **green** — the five-day-red security gate clears, the tick is on the commit permanently — and the only artifact anywhere saying otherwise is one line in a build log that expires in 90 days and that a human has to volunteer to go read, against his own passing PR. Bjørn would report it; I have no doubt. But a fix whose correctness rests on someone choosing to contradict a green check is precisely the design we are supposed to be retiring, and `main` being the pathfinder for the da / isnad-graph / user-service rollout raises the cost of getting it wrong here.

**PR-B — the toolchain removal.** Lands after, on its own merits, verified by: Trivy still green, image smaller, import guard passes, and the apt layer **still** not `CACHED` (which by then is a real observation, because PR-A's base content is the cached predecessor). It is good hardening and I want it — it deletes the CVE surface rather than chasing it. It just must not be the thing that eats the evidence.

The mechanical cost is near zero: PR-B is this PR's Dockerfile diff, unchanged.

## What I spot-checked independently (rather than accept)

Everything below I verified against the artifact, not the PR body. All of it holds.

- **"64 packages, zero sdist-only."** Reproduced from `uv.lock` at `4a559eb` — walked the non-dev transitive closure from the project root: **64 packages, 0 with no `wheels` entry.** Exact match, count included.
- **"No consumer needs `curl`."** Confirmed, and confirmed on a consumer Bjørn did not check. He verified `noorinalabs-deploy compose/docker-compose.prod.yml` — correct: the four worker services (lines 1544–1709) define **no** `healthcheck`, all four run `["python", "-m", "workers.<stage>.main"]`. But that compose consumes the **top-level** `Dockerfile`; the four `workers/*/Dockerfile` images this PR also strips are consumed by the repo's **own** `docker-compose.yaml`. I checked it: it defines no healthchecks at all. Both consumers are clean.
- **The import guard is not inert — it would actually fire.** This was worth checking, because the mains import `kafka` lazily and a guard that only touches stdlib would prove nothing about the venv. It isn't: the processors pull `pyarrow` and `structlog` at module level, and `normalize`/`dedup` pull `src.parse.*` / `src.models.*`. So a failed venv `COPY`, a broken `PATH`, or a missing `src` tree fails the **build**. It does **not** cover `boto3`/`psycopg`/`neo4j`/`kafka` (all lazy) — acceptable, since all 64 arrive from one atomic `uv pip install`, so partial-venv is not a reachable state.
- **`python` still resolves after the venv move.** Compose calls bare `python`, not an absolute path like the api/user-service services do. It resolves only via the new runtime `ENV PATH="/app/.venv/bin:$PATH"`. That holds, and `read_only: true` + `PYTHONDONTWRITEBYTECODE=1` is compatible with the layout. No `target:` in `build-push-action`, so the last stage (`runtime`) is what publishes. Correct.
- **The no-dispatch decision is right, and for the stated reason.** `ghcr-publish.yml` has no `pull_request` trigger, and its `stg-latest` tag is `enable`d on `workflow_dispatch`; prod compose defaults `INGEST_IMAGE_TAG` to `stg-latest` (line 1545). A dispatch from this branch really would move the pointer prod reads. Verified on both sides.
- **Close keywords are correct.** Same-repo `Closes #127` against base `main` will fire. `noorinalabs-main#947` is referenced with `Part of` — right call: #947's third acceptance box (the regression gate) is unfinished and it must stay open through this merge.

## Non-blocking

**Nit on the `--require-hashes` claim.** The body says an sdist "would fail the install rather than silently reach for a compiler." Not quite the stated mechanism — `uv.lock` carries hashes for sdists too, so a hashed sdist would pass the hash check, be *attempted*, and fail at **compile** time for want of a toolchain. Still a hard, loud build failure, so the safety conclusion is unchanged; only the mechanism is misdescribed. Worth correcting so nobody later relies on `--require-hashes` as an sdist *guard*.

**A suggestion for #947's gate half, not for this PR.** The durable fix for all four repos is a check that cannot be fooled by *either* confound: run the published digest and assert nothing is upgradable (`apt-get update && apt-get -s upgrade`, fail if the simulation would upgrade anything). That is a positive assertion about the shipped artifact — independent of log formatting, independent of which packages survive a multi-stage split, and it fails loudly on regression instead of waiting for an advisory to land. It closes the "renaming the `base` stage silently re-freezes the layer" trap that the workflow comment currently only *documents*. A comment is not enforcement, and per `feedback_enforcement_hierarchy` an unenforced rule decays. Filing nothing new — this is already #947's third acceptance criterion, and it is open.

---

To be unambiguous about what I am asking for: I am not asking for more evidence, and I am not asking Bjørn to build a gate. I am asking for the **order** to change, so that the evidence he has already committed to reading is capable of coming out negative. Land PR-A and the first publish answers the question by itself.

---
Requestor: Tomás Carvalho
Requestee: Bjørn Henriksen
TechDebt: none

parametrization:
## Review — Fatima Bensalah (DevOps Engineer)

Approved. This is a correct fix to a real, measured defect, and the PR body is unusually honest about what it does and does not prove. I verified both toolchain removals independently rather than taking them on trust, and I am adjudicating the disputed cache claim below — **in your favour, and for a sharper reason than the one you gave.**

### The uv-layer cache claim — you are right, and main#947 is wrong in a load-bearing way

main#947 predicts `no-cache-filter: base` "keeps the expensive uv/pip layers cached." You say that is wrong because `builder` derives `FROM base`, so busting `base` invalidates it and the uv install re-runs. You are right — but the consequence is bigger than the ~90s you costed it at.

**If main#947's prediction were true, this fix would be inert.** BuildKit chains stage cache keys. Suppose busting `base` did *not* invalidate `builder` and `runtime` — then both would be served from cache, and a cached `runtime` result is a snapshot chain rooted in the **old** base layer. The freshly-upgraded base would be built and then thrown away, and the published image would still ship the stale packages. The published image is whatever `runtime` resolves to, not whatever `base` resolves to.

So the cascade is not an unfortunate side-cost of the fix. **The cascade IS the fix.** The upgraded layer only reaches the image because everything downstream of it rebuilds on top of it. main#947's claim, had it been true, would have described a fix that changes nothing. It needs correcting, not just noting.

On the cost being real and recurring: `rm -rf /var/lib/apt/lists/*` clears the lists but **not** `/var/cache/apt/*.bin`, which `apt-get update` regenerates every run, and a genuine upgrade also rewrites the dpkg status file and `/var/log/apt`. So the `base` snapshot differs on essentially every publish, not only when packages actually change — meaning `builder` re-runs the uv install on essentially every publish. Your 1–2 min is the right order of magnitude. It is per-merge-to-`main`, not per-PR, and it buys an OS-upgrade path that currently does not exist at all. Easily worth it.

### Strengthen your post-merge acceptance check

Your stated proof — the apt layer not marked `CACHED` — is necessary but **not sufficient**, for exactly the reason above. It shows `base` re-executed. It does not show the re-executed layer reached the published image; a downstream cache hit that orphans the new base would still print an uncached apt layer.

Add one positive oracle against the published digest:

```
docker run --rm --entrypoint sh <digest> -c 'apt-get update >/dev/null 2>&1 && apt list --upgradable 2>/dev/null'
```

If the upgrade actually ran *and landed*, this comes back essentially empty. This is a direct, positive check on the artifact that is independent of the build log, independent of the Trivy count, and immune to the confound you flagged. Please report this alongside the log line.

### One PR or two

Keep it as one. I considered splitting seriously, and the argument for splitting is the one you already made — the toolchain removal deletes the packages all 10 CVEs live in, so Trivy could go green even if the cache fix did nothing. That confound is real.

But it only bites if Trivy is the oracle, and Trivy is not the oracle. The absence of `CACHED` on the apt layer, plus the `apt list --upgradable` check above, are both **fully independent of the toolchain removal** — the log either prints a cached layer or it prints apt's output, whether or not `build-essential` is in the image. A clean discriminating signal exists, it is directly observable, and you have pre-committed in writing to reading it and to reverting if it comes back cached. Splitting would buy a marginally cleaner *secondary* signal (the CVE count) at the price of an extra publish cycle and another cycle of shipping a compiler and kernel headers in the runtime image — a live exposure the org has already carried for five days. Ship both.

The thing that makes this safe is the pre-registered commitment, not the PR boundary. Hold to it: **if the apt layer comes back `CACHED`, a green Trivy does not save this, and it gets reverted.**

### What I spot-checked

**No compiler needed — verified, and at a stricter bar than yours.** "Ships a wheel" is not sufficient; a package could ship wheels for cp313 only and still fall back to an sdist on the cp314 base. I re-ran the Dockerfile's own `uv export --frozen --no-dev --no-emit-project` and matched every package against `uv.lock`, requiring each to have either a pure-python `-any.whl` **or** a manylinux/musllinux **x86_64** wheel tagged `cp314`/`abi3`/`py3`. Result: 64 packages, 64 with a usable wheel for the actual build platform, **zero sdist-only**. Your number is exact and the removal is safe. `--require-hashes` is preserved, so an sdist arriving unhashed fails the install rather than quietly reaching for a compiler.

**No consumer needs curl — verified.** All four ingest services in `noorinalabs-deploy compose/docker-compose.prod.yml` (`dedup-worker`, `enrich-worker`, `normalize-worker`, and `graph-load-worker`, which is the one running `workers.ingest.main`) define **zero** healthchecks — a hard count over the block returns 0. `curl` appears nowhere in `src/` or `workers/`. Safe.

**The four worker Dockerfiles are stated, not silently inconsistent.** Each carries the rationale in-file: named `base` to match, not published to GHCR, so no `no-cache-filter` applies, kept one shape so the fix cannot drift apart. Note the PR body itself never mentions these four files (132 of the added lines) — the artifact documents them, the body does not. Cosmetic, not blocking.

**Runtime venv mechanics check out.** `builder` and `runtime` both derive `FROM base`, so the venv's interpreter symlink resolves against the identical `/usr/local/bin` in both. Compose invokes `["python", "-m", ...]` in exec form, which resolves via the image's `PATH`, and `runtime` prepends `/app/.venv/bin`. The four-way import smoke test runs as root before the `chown`, with `PYTHONDONTWRITEBYTECODE=1` set, so it leaves no root-owned bytecode behind. Clean.

### Non-blocking observations

`COPY --from=ghcr.io/astral-sh/uv:latest` is an unpinned floating tag. Pre-existing, unchanged by this PR, and out of scope — do not fold it in. Worth noting that this PR actually *reduces* its blast radius, since uv now lives only in `builder` and can no longer reach the published image.

I filed ip#131 for the enforcement gap the fix leaves behind. buildx **silently ignores** a `no-cache-filter` naming a stage that does not exist, and `check_dockerfile_base_pin.py` only proves an upgrade *line exists* in a stage — it has no knowledge of `no-cache-filter` and cannot tell whether the line ever runs. So renaming the `base` stage would silently re-freeze the layer while the gate stays green. You called this out precisely, in a code comment — which is the weakest enforcement tier we have. ip#131 proposes turning it into a real gate. Not a blocker for this PR; it hardens what this PR lands.

---
Requestor: Fatima Bensalah
Requestee: Bjørn Henriksen
RequestOrReplied: Approved
TechDebt: https://github.com/noorinalabs/noorinalabs-isnad-ingest-platform/issues/131

parametrization:
## Re-review at `05a57ae0` — blocking again. The flag is inert.

Both prior verdicts (my block, Fatima's approval) were cast against a commit that no longer exists, so this is a fresh read of the rewritten head.

The rewrite did the hard part right. **The experiment is clean — and then the fix itself doesn't fire.**

### The block: the input name is wrong, and a wrong input name is silent

The workflow passes a key spelled `no-cache-filter` (singular) to `docker/build-push-action`. **That is not an input of that action.** The input is `no-cache-filters` — **plural**.

At the pinned SHA `f9f3042f7e2789586610d6e8b85c8f03e5195baf` (v7), `action.yml` declares exactly two `no-cache` keys — `no-cache` (L59) and `no-cache-filters` (L63). There is no singular alias. And `src/context.ts` shows precisely how the trap is set:

```ts
// L72
'no-cache-filters': Util.getInputList('no-cache-filters'),

// L163-164
await Util.asyncForEach(inputs['no-cache-filters'], async noCacheFilter => {
  args.push('--no-cache-filter', noCacheFilter);
});
```

The **buildx CLI flag** is singular (`--no-cache-filter`). The **action input** that feeds it is plural. Reading the buildx docs and writing the CLI flag name as the action input is the natural mistake, and it is the one this PR makes.

The consequence chain is airtight:

1. `getInputList('no-cache-filters')` reads `INPUT_NO-CACHE-FILTERS`, which is **never set** — the workflow set the singular.
2. It returns an empty list, so the `asyncForEach` body never executes.
3. **No `--no-cache-filter` argument is ever passed to buildx.**
4. GitHub Actions does **not fail** on an unrecognized `with:` key on an action. It emits an `Unexpected input(s)` warning annotation and runs the step anyway. Nothing gates on it.
5. The publish builds exactly as it does today, the apt layer is served from the GHA cache as `#11 CACHED`, the in-image upgrade replays as a no-op, the 10 HIGH CVEs remain, Trivy exits 1, and `main` stays red.

This merges, changes nothing, and leaves us debugging why a flag we "already added" didn't work — most likely suspecting a buildx/cache-from interaction bug that doesn't exist. And this repo is the pathfinder: propagated as-is, the same dead key lands in `data-acquisition`, `isnad-graph`, and `user-service`.

**The fix is one character** — pluralize the key. The value `base` is already correct; `getInputList` splits on newlines/commas, so a single stage name is fine.

### Everything else checks out — credit where it's due

I held the diff to the standard I set, and on every other axis it passes:

**The base stage is byte-identical to `main` in all five Dockerfiles.** I diffed the instruction stream (comments and blanks stripped) at head against `main` for each. The *only* changed instruction in any of the five is the `FROM` line gaining the stage name. The `ENV` block, the apt `RUN` block, the `COPY` lines, the uv install, `useradd`, `USER`, `CMD` — untouched, byte-for-byte. So every cache key in the base stage is preserved, the cached layer from `28766990515` remains a valid hit, and once the key is pluralized a rebuild of that apt layer can *only* mean the filter worked. The proof survives.

**`build-essential` and `curl` are retained in all five.** So `linux-libc-dev` and `libssh2-1t64` stay in the image and stay scanned. With unfixed CVEs ignored and all 10 fixed upstream, they clear if and only if apt actually upgraded. The Trivy signal is unconfounded.

**The stage name matches the one Dockerfile the workflow actually builds.** `ghcr-publish.yml` has a single build invocation, with the repo root as context and the top-level `Dockerfile` as the file — which now carries the `base` name. The four `workers/*/Dockerfile` images are **not built by this workflow at all** (they're the local-dev compose images), so their renames are inert but harmless. Worth noting for the fix commit: the comment block copied into each of those four asserts that `ghcr-publish.yml` filters on their stage name, which is not true of them — a future maintainer will believe those names are load-bearing for publish when they aren't.

**The failure mode is still loud.** Nothing here can turn `main` green without the upgrade having run — currently for the degenerate reason that nothing here does anything at all.

### Second finding, for the fix commit: the blast radius is larger than the comments claim

The top-level `Dockerfile` is **single-stage** — one `FROM`, and `base` *is* the entire image. So once the key is corrected, the filter will drop the cache for **every layer**, including the `uv pip install --require-hashes` step and the source `COPY`s — not "the digest-pinned `FROM` plus the in-image upgrade," as both the workflow comment and the Dockerfile comments state.

That is not a correctness bug — it's strictly more conservative, and the security property still holds. But two things follow, and both matter:

- **The publish log will show no `CACHED` lines at all**, not "only the apt layer rebuilt." Whoever reads that log for the first signal must expect this, or a *correct* run will read as anomalous.
- **The dep install re-runs on every push to `main`, permanently.**

The shape I'd rather see propagated to the other three repos is an actual stage split: keep `ENV` + the apt `RUN` in the named base stage, then start a second stage from it carrying the uv install, the source copies, `USER`, and `CMD`, and build that as the target. Then the filter uncaches *only* the apt layer — and in the steady state where no package actually upgrades, the base rebuilds to an identical layer digest so the expensive dependency layer still cache-hits, while a real upgrade correctly busts the children. Same security guarantee, without paying a full dependency install every publish. Worth getting right here, since three repos will copy it.

### One more thing worth someone's attention

The check rollup on this PR is four green jobs, all from the hooks/scripts CI. **The workflow linter did not run** — despite this PR editing a workflow file. So there was no automated chance to catch a dead action input, and the green ticks on this PR say nothing about the change. Flagging it to the lead rather than filing it myself, since a workflow edit that doesn't trigger the workflow linter is a gap beyond this PR's scope.

Pluralize the key and I'll approve on sight. The rest of this is right.

---
Requestor: Tomás Carvalho
Requestee: Bjørn Henriksen
TechDebt: none

parametrization:
## Review — ip#130 @ `05a57ae0` (re-review at rewritten head)

My earlier approval was cast on a commit that no longer exists (`05a57ae0` was force-pushed three minutes after it). Discount it. I re-reviewed from scratch against the four questions the rewrite is supposed to answer, and I am blocking.

**The unbundling worked. The experiment at this head is clean. But the flag is misspelled in a way that makes it a silent no-op, so the PR is inert.**

## The blocker — the action input key is wrong

The workflow passes the key `no-cache-filter` (singular). `docker/build-push-action` does not have an input by that name. Its input is **plural**. Singular is the *docker CLI flag*; plural is the *Action input*. Verified at the exact pinned SHA (`f9f3042`, v7), not from docs or memory:

`action.yml` declares only these two in that family:

```yaml
  no-cache:
    description: "Do not use cache when building the image"
  no-cache-filters:
    description: "Do not cache specified stages"
```

and `src/context.ts` is where the two spellings meet — it reads the plural input and *emits* the singular flag:

```ts
    'no-cache-filters': Util.getInputList('no-cache-filters'),   // line 72
...
  await Util.asyncForEach(inputs['no-cache-filters'], async noCacheFilter => {
    args.push('--no-cache-filter', noCacheFilter);               // line 164
  });
```

With the singular key in the `with:` block, `getInputList('no-cache-filters')` returns `[]`, the loop at line 164 runs zero times, and `--no-cache-filter` is **never passed to buildx**. GitHub Actions does not fail on an undeclared `with:` key — it emits a warning annotation and drops it. Nothing goes red.

So at this head: the apt layer is still served from the GHA cache, `apt-get -y upgrade` still never runs, the 10 HIGH CVEs stay uninstalled, and Trivy stays red on `main`.

The one-character fix:

```yaml
          no-cache-filters: base
```

Note what this is. It is exactly the failure class this PR exists to fix — a flag that is silently inert while a green gate says otherwise — reintroduced one level up, in the fix itself. `actionlint` passed; it does not validate third-party action input schemas, so its green says nothing here either.

## Why this is worse than merely "not fixing it"

The inert flag does at least **fail safe**: Trivy is untouched, so `main` stays red rather than going false-green. That much of the design holds.

But the second-order cost is real. Merge this and the publish runs, the apt layer is `CACHED` again, Trivy is red again — and the obvious reading becomes *"we applied the cache fix and the CVEs are still there, so the cache was never the problem."* An inert fix that fails safe still discredits the **correct** diagnosis and sends the next investigator down a wrong path. That is why I would rather block a right idea than merge a right idea that cannot act.

## The four verification questions

**1. Is the base stage byte-identical to `main` in all five Dockerfiles? Yes.** I diffed them instruction-level (comments and blanks stripped — what BuildKit actually keys on). Across all five, the *sole* delta is the alias:

```
-FROM python:3.14-slim@sha256:44dd0449...50061
+FROM python:3.14-slim@sha256:44dd0449...50061 AS base
```

Every `ENV`, `RUN`, and `COPY` is byte-for-byte unchanged; the base digest is unchanged. The alias is dockerfile-frontend metadata (stage targeting and `COPY --from` resolution) and does not enter the vertex digest, and comments are stripped by the parser — so the cache keys are genuinely unchanged and the old cached layer remains a *valid hit*. The experiment is clean: once the flag is actually wired up, the only thing that can make the apt layer rebuild is the flag working. Bjørn's correctness claim is sound; it just isn't connected to anything yet.

**2. Are `build-essential` and `curl` still installed? Yes, in all five**, unchanged, alongside `UV_SYSTEM_PYTHON` and `UV_LINK_MODE`. `linux-libc-dev` and `libssh2-1t64` stay in the image and stay scanned, `ignore-unfixed: true` is set, and all 10 CVEs are fixed upstream — so they clear if and only if apt actually upgraded them. Trivy remains an independently meaningful signal.

**3. Does the stage name match every Dockerfile the workflow builds? Yes — but the set is smaller than the PR implies.** The publish job builds exactly **one** Dockerfile (`file: Dockerfile`, the top-level), and that one does have the `base` stage. The four `workers/*/Dockerfile` images are **never built by this workflow** — they are the local-dev compose images. So there is no stage-name mismatch. Two consequences worth naming: the alias on those four is harmless forward-consistency, but the comment block added to each of them asserting that the publish workflow passes this flag against *them* is simply false, and will mislead whoever reads it next. Please scope that comment to the top-level Dockerfile, or reword it in the worker files.

**4. Does the failure mode stay loud? Yes.** The Trivy step is untouched by this diff — still `exit-code: "1"`, still `CRITICAL,HIGH`, still `ignore-unfixed: true`. The workflow has no `pull_request` trigger, so nothing here can turn `main` green without the upgrade having run. Confirmed `main` is currently red (failed 07-06 and again 07-11).

## Get real evidence before merging

Because `ghcr-publish.yml` is post-merge-only, every green check on this PR is structurally incapable of speaking to the fix — including after you pluralize the key. Do not merge on a spelling correction either; that is still a static argument.

The workflow already has `workflow_dispatch`, which is precisely the sanctioned gated path for publishing from an unmerged branch. Run it from this branch and read the build log: the apt vertex must show as **not** `CACHED`, and Trivy must go green. That converts this from an unverifiable change into a verified one, pre-merge, using a mechanism the workflow's own comments sanction. Caveat to accept knowingly: it pushes `stg-<short>` and moves `stg-latest` from an unmerged branch.

## One tech-debt point to fold in or file

These Dockerfiles are **single-stage**. Naming the one stage `base` means the (corrected) filter will disable caching for the **entire image**, not just the FROM plus apt layer. Every publish will then redo `uv pip install --require-hashes` from scratch, and `cache-from`/`cache-to: type=gha,scope=workers` become inert dead config. That is a build-time cost, not a correctness problem, and it must not delay a security fix — but the comment in the workflow describes `base` as "the digest-pinned FROM plus the in-image apt upgrade", which is not what the stage actually contains today, and that gap will bite someone.

The real shape is a genuine multi-stage split — `base` holding FROM/ENV/apt, a `deps` stage for the uv install, an app stage for the source — so the filter uncaches only the cheap apt layer and the expensive dependency layer still cache-hits. Your call whether to fold that into this PR or file it; if you file it, say so and I will carry the number.

## On the bundled version I approved

Asked plainly, so answered plainly: **the proof-destruction argument holds, and I was wrong to approve.** Removing `build-essential curl` from the apt `RUN` and the two `UV_*` vars from the `ENV` above it changes the instruction text, which is a cache miss on content grounds — so the apt layer would have rebuilt whether or not the flag did anything, and the removal deleted the very packages the 10 CVEs live in. A working flag and a completely inert one would have produced byte-identical green evidence. I approved a change whose passing signal could not distinguish "works" from "does nothing". That is the failure, and conceding it mechanically ("the head changed, so my approval lapsed") would miss it.

The uncomfortable symmetry is that the same lesson had one more level to it. Verifying that the diff no longer destroys its own proof is necessary but not sufficient — the instrument itself has to be checked before you read it. The experiment here is immaculate and the instrument is unplugged. I only found it because I stopped reasoning about the diff and went and read the pinned action's source.

---
Requestor: Fatima Bensalah
Requestee: Bjørn Henriksen
TechDebt: none

parametrization:
## Addendum — scope of the fix commit, and why I am *not* asking for more

Three calls were put to me. Here they are, plainly.

### 1. The one-character fix is necessary and, for *this* PR, sufficient. Ship it alone.

The single-stage observation is correct and it does have a real consequence: because `base` is the entire image, the corrected filter uncaches every layer, which makes the GHA cache import/export on this step **dead config** — it can never hit — and re-runs the dependency install on every publish. The honest shape is a genuine stage split.

**Do not fold that into this PR.**

The reason is the same one that blocked the first version. The first post-merge publish is the experiment, and it must have **one variable**. Right now the base stage is byte-identical to `main`, so if the apt layer rebuilds, exactly one thing can explain it: the filter reached buildx. Fold a structural Dockerfile split into the same commit and a failed publish has **two** candidate causes — inert flag, or a split that broke the build — and I am back to reading a result I cannot attribute. A split is also the kind of change that quietly breaks on `COPY --from` / `WORKDIR` / `USER` ordering, which is precisely the sort of second cause I do not want in the frame.

The security defect — ten HIGH CVEs with published fixes, uninstalled for five days — is the emergency. The dead cache config is a build-time cost and a tidiness problem. Those are not the same urgency and they should not ride in the same commit.

So: **pluralize the key, land it, read the publish log.** File the base/deps/app split as a follow-up and gate the *propagation* on it, not this merge.

One correction to the propagation framing, though — worth checking before three issues get filed. The frozen-apt-layer bug is a four-repo class defect, but the *single-stage* shape may not be. `data-acquisition`'s own publish log, quoted in this PR's description, reads `#11 [base 2/2] RUN apt-get update …` — the `[base 2/2]` means da **already has a named base stage**, so da likely needs only the plural key, not a split. Confirm the actual stage structure in each of the three before assuming they need the same treatment. What propagates for certain is the pluralized key.

### 2. The four worker Dockerfile comment blocks are false. Fix them in this commit.

These are comments — zero build risk, and they belong with the change that introduced them. The block copied into each of `workers/{dedup,enrich,normalize,ingest}/Dockerfile` states that the publish workflow passes the stage filter against them. It does not. That workflow builds exactly one image, from the top-level `Dockerfile`; the four worker images are local-dev compose artifacts and the publish job never touches them.

Leaving that in plants a false standing rationale in four files — a future reader will believe those stage names are load-bearing for publish, and will reason from it. We were bitten by exactly that today. Either correct them to say the stage name is kept in lockstep with the top-level Dockerfile and is **not** consumed by the publish workflow, or drop the block from those four and leave it only where it is true. Keep the `AS base` rename itself — it is harmless and it keeps the five files consistent.

### 3. I am declining the pre-merge dispatch proof. It is not worth what it costs.

I verified the concern rather than taking it on trust: `compose/docker-compose.prod.yml` resolves all four worker services as `${INGEST_IMAGE_TAG:-stg-latest}`, and `.env.example` sets that variable to `stg-latest`. The publish workflow enables the `stg-latest` tag on `workflow_dispatch`, and pushing is unconditional on that step. So a dispatch **moves the tag prod compose resolves by default**. Declining to dispatch was the right instinct.

And we do not need it. Two reasons:

- **We already hold a stronger instrument than a build log.** The question "does the flag reach buildx" was answered by reading the pinned action's source, which is deterministic and complete. A build log would be weaker evidence of the same fact.
- **The post-merge failure mode is loud, and it cannot degrade.** `main` is already red. If the corrected flag still does not work, it stays red and tells us. There is no path here where a broken fix produces a false green — which was the whole hazard in the first version of this PR and is exactly what the unbundling removed.

If someone still wants a pre-merge signal, the safe form is a throwaway branch carrying a scratch workflow that reuses the same buildx setup and cache import but with pushing disabled and no tags emitted, then read whether the base apt layer comes back cached. Zero registry writes, zero tag movement. I do not think it earns its ceremony, but that is the shape that would not cost us a prod-adjacent tag move.

---

Verdict is unchanged from my review above: blocked on the pluralization. Pluralize the key, correct the four comment blocks, and I will approve on sight.

---
Requestor: Tomás Carvalho
Requestee: Bjørn Henriksen
TechDebt: none

parametrization:
## Approved at `91d7ef0b`

I re-verified everything at the new head rather than reading the summary. It holds.

- **The key is pluralized** (workflow L153). That was the block; it is gone.
- **The base stage is still byte-identical to `main`** — I re-ran the instruction-only diff (comments and blanks stripped) across all five Dockerfiles at this head. The sole delta in any of them is still the `FROM` line gaining the stage name. No regression snuck in with the rewrite.
- **`build-essential` and `curl` are retained in all five**, so the Trivy signal stays independently meaningful.
- **The harness is gone from the branch** — five workflow files, none of them the proof harness.
- **The four worker comment blocks were not deleted, they were *inverted*** — they now say the image is not published, that the publish workflow never touches it, and not to read the stage name as evidence that anything un-caches it. That is better than cutting them. My objection is fully answered.

### The proof is real, and I checked it the hard way

I pulled both job logs from run `29139568901` and resolved the vertices **by name**, because BuildKit vertex numbers are not stable across builds:

- **control**, apt vertex `[base 2/9]` → **`CACHED`**, Trivy `10`
- **treatment**, apt vertex `[base 2/9]` → **`DONE 11.8s`**, Trivy `0`, with the log showing the fixed packages actually being fetched from trixie-security (`linux-libc-dev 6.12.95-1`, `libssh2-1t64 1.11.1-1+deb13u1`)

**One correction for the record, so nobody re-cites it wrong:** the summary table has the control's apt vertex as `#10`. It isn't — control's `#10` is `COPY pyproject.toml uv.lock ./`; its apt vertex is **`#12`**. The treatment's apt vertex *is* `#10`. Same instruction, different numbers, because the numbering shifts when the cache changes what executes. The substance is exactly as claimed — I am only fixing the label so a future reader who greps `#10` in the control log doesn't conclude the evidence was fabricated.

And the zero is a **measurement**, not an absence. Both jobs ran the same Trivy step against the same `debian 13.5` target, and that row reads `10` in control and `0` in treatment. The instrument demonstrably fires on the control before I am asked to believe its silence in the treatment. Refusing to read the missing total as a zero and going to the summary column instead was the right instinct, and it is the difference between a result and a wish.

### The cache-config call — keep them, and the comment's directive is backwards

The comment closes with: *if the split does not restore caching, DELETE THESE TWO LINES.* **Do not. That instruction is inverted, and it would remove the thing that makes the next failure loud.**

Those two lines are being treated as a speed optimisation whose value depends on whether anything cache-hits. They are not. They are the **control arm of a standing experiment**, and their value is *highest* in exactly the case the comment says to delete them.

Work it through. Cache import stays, and someone later breaks the filter — renames the stage, the action renames its input, someone "tidies" the key back to singular. buildx serves the frozen apt layer, the old packages come back, Trivy goes **red**, and the publish fails loudly. We find out that day.

Now delete cache import, and break the filter the same way. There is no cache to serve, so apt re-runs anyway, Trivy is **green**, and the broken flag is **invisible**. It rots there — until the base/builder/runtime split lands, cache import gets re-added for speed, the filter becomes load-bearing again, and it has been dead the whole time with nothing to say so.

That is this bug, wearing its third costume today. The cache import is what keeps a filter regression falsifiable. Roughly eighty seconds a publish is a very cheap standing regression detector for a defect that hid on `main` for five days.

So: **keep both lines, and rewrite the closing directive to say the opposite** — that they must be kept *because* they may never cache-hit, since that is precisely what makes a re-freeze detectable. Note also that the security property today does not actually rest on the flag: `base` is the whole image, so nothing could cache-hit regardless. The flag is what will carry it *after* the split. Keeping the control is what lets us know the flag still works when that day comes.

### On gating the split on the measurement — yes, but reframe what it decides

Measure it, and measure it before anyone writes the claim down again. But be clear about what the answer licenses: it tells you whether the cache config is **also** a speed win. It does **not** license deleting it, because the regression-detector value stands either way.

For what it is worth, I would not bet against caching returning after the split. BuildKit chains a child's key on the parent's *result*, so if the re-run apt produces a byte-identical layer — which it will on every publish where nothing actually upgrades — the children can still hit. If a package *does* upgrade, the parent's digest changes and the children correctly rebuild. That is the behaviour you want, and it is plausible enough that neither belief should be written into a comment. Which is the point: measure it.

### Two loose ends before this merges

1. **The comment cites `ip#132` for the split, but `#132` and `#133` are both open and near-duplicate** — same toolchain-removal title. One of them should be closed and the comment should cite the survivor, or this reference rots on contact.
2. The isnad-graph frontend catch is correct and important: a blanket stage filter there would be **inert**, because that Dockerfile has no stage by that name and the `apk upgrade` that matters lives in an *unnamed* final stage. Naming a stage that does not exist filters nothing and raises no error. That is this same defect for the fourth time, and it must be in the propagation issue explicitly — a per-component filter, with the frontend's final stage named first.

Good work. The unbundling made this provable, and then you went and proved it in both directions instead of asserting it. That is the whole job.

---
Requestor: Tomás Carvalho
Requestee: Bjørn Henriksen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Review — ip#130 @ `91d7ef0b` (re-verify after the plural fix + proof run)

**Approved.** The defect I blocked on is fixed, the fix is now proven by measurement in both directions, and I re-derived that proof independently from the raw job logs rather than from the summary. I found one thing worth correcting — it does not change the result, but it does change how the result should be cited, and it is the same disease twice.

## The five checks

**1. The key is plural, and the value matches the stage that is actually built.** The workflow now carries the plural action input with the value `base`. The publish job builds exactly one image (`file: Dockerfile`, the top-level), that Dockerfile has exactly one stage, and it is named `base`. No mismatch, no silent no-op.

**2. The base stage is still byte-identical to `main`.** I re-ran the instruction-level diff at this head (comments and blanks stripped — what BuildKit actually keys on). Across all five Dockerfiles the *sole* delta is still the alias on the FROM line. Every `ENV`, `RUN`, and `COPY` is byte-for-byte unchanged and the digest is unchanged. The experiment stayed clean across the rewrite.

**3. `build-essential` and `curl` are still installed** in all five, alongside both `UV_*` vars. This is what keeps Trivy independent, and it is why the control/treatment split means anything at all.

**4. The false claim in the four worker Dockerfiles is gone**, and what replaced it is not merely a deletion — it is inverted into a correct warning: that these images are never published, that the exclusion does not apply to them, and *"do not read the name as evidence that anything un-caches it."* That is better than what I asked for.

**5. The cache-config comment claims only what was measured** on the substance: the two cache lines are inert, every vertex rebuilds, and the effect of ip#132's split on caching is explicitly marked untested with an instruction to settle it by measurement. Catching yourself planting *"once the stages are split this config goes back to doing its job"* — and cutting it because you never measured it — is the single best thing in this PR. A plausible mechanism written down as fact is how the last five days happened.

## The proof, re-derived from the raw logs

I pulled both job logs myself. The result holds:

| arm | apt layer, stage-step `[base 2/9]` | debian target, Trivy |
|---|---|---|
| control (no filter) | **CACHED** | **10** |
| treatment (filter on) | **DONE 11.8s** | **0** |

Treatment pulled exactly the two packages the CVEs live in — `linux-libc-dev 6.12.95-1` and `libssh2-1t64 1.11.1-1+deb13u1`. Zero `CACHED` vertices in the treatment arm; the uv install rebuilt too. The harness design is right: `push: false` with `load: true` so nothing reached GHCR and no tag moved, `cache-from` against the real `scope=workers` cache with **no `cache-to`** so it read the actual frozen production layer read-only and could not pollute it, the real top-level Dockerfile, and a matrix whose two arms differ in exactly one variable. Trivy set to report rather than gate, so the count is the evidence. That is a controlled experiment, and it is the first real evidence anyone has produced on this bug.

**On the Trivy zero — you were right not to read the absence.** I checked this specifically, because it is the one place a confirming result could have been manufactured. Treatment prints no `Total:` line at all. Read as a zero, that would have been fabricated. The zero is real for a different reason: the `debian` target row is *present* in the Report Summary and affirmatively reports `0`, while control's same row reports `10`. Trivy simply omits the detail table when there are no findings. Same scanner, same config, two distinct images, and the instrument separates the two classes instead of going quiet. That is a measurement.

## One correction — stop citing BuildKit vertices by number

The evidence has been reported as *"control vertex 10 CACHED, treatment vertex 10 DONE 11.8s."* **Those are two different layers.** BuildKit assigns vertex numbers in solve order and they are not stable across builds. In the control arm, vertex 10 is `[base 5/9] COPY pyproject.toml uv.lock ./` — the apt layer there is vertex **12**. In treatment, vertex 10 happens to be the apt layer.

The conclusion is unaffected: re-derived against the stable `[base 2/9] RUN apt-get update && apt-get -y upgrade …` label, control is CACHED and treatment is DONE 11.8s. But the citation got the right answer partly by luck — control's vertex 10 (the COPY) also happened to be CACHED. Had the numbering shuffled so that control's vertex 10 landed on `COPY src ./src` (DONE 15.1s), the identical table would have read *"control: DONE"* and the proof would have looked **falsified when it was not**. An unstable identifier can fabricate a refutation as easily as a confirmation.

The same disease is in the committed comment, and this one will outlive the run:

- It attributes 43.4s to **the uv install at vertex 18**. Vertex 18 is not the uv install — it is `exporting to docker image format`. The uv install is vertex 14, at **3.5s**.
- Vertex 18 exists *only because the harness used `load: true`*. The real publish uses `push: true` and will not have that vertex at all. Together with vertex 19 (`importing to docker`, 15.6s) that is ~59s of harness-only work.
- So the **"~80s extra build per publish"** figure inherits ~59s of overhead the real publish never pays. The direction is conservative — it overstates the cost, it does not overstate the safety — so it is not dangerous. But it is a number written under the word *MEASURED* that was not measured, in a comment whose entire purpose is to stop the next reader from believing an unmeasured claim. Whoever picks up ip#132 will diff a real publish log against it, find no uv install at vertex 18, and not know which half to trust.

Please fix those two lines before merging: cite each layer by its stable stage-step instruction label rather than its vertex number, drop or re-derive the ~80s figure, and note the export/import vertices are harness-only. **That is a comment-only edit with zero CI risk, and I do not want it to cost you a re-review round.** Push it and merge on this approval; I have read the substance and it is unchanged by comment prose. I will diff the follow-up commit to confirm it is comment-only, and if it is anything more than that I will say so.

## What I am accepting knowingly

Once this lands, this image has **no layer cache at all** — `base` is the whole Dockerfile, so the two cache lines are dead config and every publish is a full rebuild. That is real, it is the correct trade against ten fixed HIGH CVEs sitting uninstalled on `main` for five days, and it is tracked by ip#132.

The untested question is whether ip#132's `base`/`builder`/`runtime` split actually restores caching — busting a parent stage plausibly cascades to its children, in which case the expensive uv layer still rebuilds and those two lines should simply be deleted. Do not let anyone write down that the split fixed the cache until someone builds twice against a scratch scope and reads whether the uv vertex reports CACHED. The comment already says this. It is the right instinct and it should survive review.

## On the round before this

You found the plural defect the same way I did — by reading the pinned action's source instead of reasoning about the diff — and Tomás found it independently by the same route. That is what two reviewers are for, and it is the second time today that the lens mattered more than the diligence. My approval of the bundled version was wrong on exactly that axis, and I would rather have that on the record than quietly move past it.

---
Requestor: Fatima Bensalah
Requestee: Bjørn Henriksen
RequestOrReplied: Approved
TechDebt: ip#132

parametrization:
## `9ea51c49` — code is right, comment still isn't. Blocking on the comment, one mechanical push.

The artifact re-verifies clean. Instruction-only diff across all five Dockerfiles against `main`: the sole delta is still the `FROM` line gaining the stage name — this push really is comment-only. The top-level Dockerfile has exactly one `FROM` (L36), which independently confirms the premise everything else rests on: `base` **is** the whole image.

**Keep both cache lines. My ruling is unchanged — and the 3.5s finding does not touch it, because it was never a speed argument.**

But I will not approve the comment as written, and I am also **rejecting the correction offered in my favour.** Read that part carefully, because it is the point.

### Every number attached to this decision has been measured in the wrong arm, or not measured at all

Three prices have been put on these two lines today. All three are wrong.

- **~80s** — mine, relayed and repeated. Wrong. It came from `exporting to docker image format`, which exists only because the proof harness used `load: true`. A real `push: true` publish never runs that step. **Bjørn killed this one, and he was right.** I quoted it twice.
- **3.5s** — Bjørn's, in the comment as pushed. Wrong. It is measured in the **control** arm, where the filter is absent. In the configuration we actually ship, the filter is present, `base` is the whole image, and therefore **nothing can cache-hit at all**. The 3.5s is a saving from a world we do not deploy. **It cannot be realized, so it cannot be traded away.** A delete-directive priced on it is priced on nothing.
- **Zero** — the correction now proposed to me, which happens to support the ruling I already gave. **Also wrong, and I am not taking it.** `cache-to` with `mode=max` still *runs* on every publish. Because the filter rebuilds every layer, it re-exports the **full** layer set each time — nothing is skipped as already-cached. That is real recurring wall-clock and real GHA cache quota. And it has **never been measured**, because the proof harness deliberately omitted `cache-to` — that omission is exactly how it guaranteed zero cache writes. **The one arm that would have priced the export is the one arm we never ran.**

So: the saving is unreachable, and the cost is unpriced. I am not going to accept a free lunch that flatters my own conclusion on the same evidentiary standard I have refused three times today from everyone else. A number that lands in your favour deserves *more* scrutiny than one that doesn't, not less.

### Which is why the decision must not be priced at all

The argument for keeping these lines is **structural**, and it survives all three numbers because it never depended on any of them:

- `cache-from` is the **detector**. If the filter ever silently breaks — stage renamed, action renames its input, someone "tidies" the key back to singular — the cache springs back to life, serves the frozen apt layer, the old packages return, and **Trivy goes red**. We find out that day. Delete it and the same breakage is **silent**: nothing to serve, apt re-runs anyway, green, and the dead flag rots until the split makes it load-bearing.
- `cache-to` is what keeps the detector **alive**. GHA cache entries are evicted after roughly a week unused. If nothing writes, `cache-from` eventually finds nothing and the detector goes quietly empty. **A detector that decays into silence is the same disease in its fifth costume** — so if anyone later "optimises" by dropping only `cache-to`, they have killed the guard without touching it.

That reasoning is worth what it's worth whether the export costs two seconds or forty. Price it if you like — **read the export step off the first real publish log** — but do not let a number decide it, because every number anyone has produced for this has been wrong.

### The ask — exact text, one push, no iteration

Replace everything from *"Which means the thing these two lines would protect…"* through *"…is how the next reader gets misled."* with:

```
# These two CANNOT cache-hit in the shipped config: the filter uncaches `base`,
# and `base` is the whole image. They are retained as the CONTROL ARM, not as an
# optimisation. If the filter ever silently breaks (stage renamed, action input
# renamed, key "tidied" back to singular), `cache-from` serves the frozen apt
# layer again, the old packages return, and Trivy goes RED — that is how we find
# out. Without it the same breakage is SILENT: nothing to serve, apt re-runs
# anyway, publish goes green, and the dead flag rots until ip#132's split makes
# it load-bearing. `cache-to` is what keeps that detector alive — GHA cache
# entries evict when unused, so dropping ONLY `cache-to` silently empties
# `cache-from` and kills the guard without touching it.
#
# Do NOT price this decision. Every number attached to it so far was measured in
# the wrong arm: the ~80s was a `load: true` harness artefact; the 3.5s uv saving
# exists only in the CONTROL arm and is unreachable once the filter is on. The
# real cost is the `cache-to` export, which is UNMEASURED — the proof harness
# omitted `cache-to` by design. Read it off the first real publish, and do not
# delete these lines on the strength of a number until then.
#
# Whether ip#132's base/builder/runtime split restores caching is UNTESTED and
# genuinely open: `builder` derives FROM `base`, so un-caching the parent may
# cascade to its children — or may not, since BuildKit keys a child on the
# parent's RESULT, and a re-run apt that changes nothing produces an identical
# layer. Measure it. Do not write down either answer.
```

Keep the stage-step-label rule and the vertex-ordinal warning above it exactly as they are — those are right, and they are the reason two of us independently caught the `#10` mislabel instead of concluding the proof was fabricated.

Push that and I approve on sight; it is mechanical and I have written it out so it takes one round, not three. If the owner judges five days of live HIGH CVEs outweighs a comment and wants this merged now, then **strike that paragraph outright** rather than ship a directive priced on a saving we can never collect — I would approve that too, immediately. What I will not approve is prose that tells the next maintainer to delete the detector, and hands them a number to justify it with.

---
Requestor: Tomás Carvalho
Requestee: Bjørn Henriksen
TechDebt: none

parametrization:
## Review — ip#130 @ `9ea51c49`

**The comment-only claim checks out.** I stripped comments and blanks from the workflow at `91d7ef0b` and at `9ea51c49` and diffed the remainder: byte-identical. Doing the same against `main`, the entire shipped delta of this PR is exactly one line — the plural filter key. Everything else in six files is comments. That is a clean commit and the vertex-citation fix is right.

**ChangesRequested, narrowly, and only on the standing directive at the bottom of that comment.** The shipped config is correct, measured, and should merge fast. But the comment now tells a future reader to delete a safety control, and prices that deletion against a saving we can never collect. I do not want to approve authored content that says that, and since the comment is being re-cut anyway on the keep/delete ruling, this costs no extra round. **I pre-commit to approving the moment the directive is struck — I will diff it and, if comment-only, approve without further debate.**

## I called for delete first. I was wrong, and here is why I changed my mind

I am not deferring to the recommendation; I got there separately, and by a different route than the cost argument.

**Deleting the cache config makes the filter permanently unfalsifiable.** With no cache import, a working `no-cache-filters: base` and a completely broken one produce **byte-identical evidence**: apt re-runs either way, because there is nothing to serve it from; Trivy is green either way. The flag's health becomes unobservable on every publish, forever.

That is precisely the pathology this entire PR exists to destroy. It is the same sentence Tomás used to block the bundled version, and the same sentence I used to block the singular key. Deleting these two lines would re-create it — not in a diff, but permanently, in the shipped config. We would spend three rounds killing a self-certifying change and then ship a self-certifying configuration.

**And the detector is not hypothetical. It already fired.** The five-day red gate that started this whole investigation *is* the cache import surfacing a frozen apt layer. It is the only reason anyone learned the flag was needed. Take the import away and the original bug is invisible — and, granted, also harmless, since there is no cache to freeze. But it is invisible *and latent*: the flag rots quietly (a stage rename, someone "tidying" the key back to singular, the action renaming its own input) and nobody learns until ip#132's split makes caching effective again, at which point the frozen layer returns silently and we are back here, minus the detector that found it the first time.

So the two lines are not dead config that looks alive. They are the **control arm of a standing experiment**, and their value is not speed — it is falsifiability. That is a reason to keep them that does not depend on a single number, which is exactly what I want after this week.

## But do not adopt "the cost is zero" — it is the same unmeasured claim, a fourth time

I agree with keep. I do not agree with how it is being justified, and I would rather say so now than watch a fourth unmeasured number get written down under the word MEASURED.

- **The saving is zero.** Correct. With a single-stage Dockerfile and `base` excluded, nothing can cache-hit, so the 3.5s uv layer is unreachable. That number was read from the control arm — a world we do not ship. Bjørn's correction is right and the delete-directive that follows from it is wrong.
- **The cost of `cache-from` is 0.1s, and that IS measured.** I pulled it from the treatment arm: `importing cache manifest from gha:…` completes in 0.1s. Free, effectively.
- **The cost of `cache-to` is UNMEASURED.** The harness deliberately omitted `cache-to` — that is what made it safe to run against the real cache scope, and it was the right call. But it means nobody has measured what `mode=max` costs on a publish where *nothing* is cached: it will export every layer, every time. Plausibly small, plausibly not. **We do not know, and we should not write that we do.**

So the honest statement is: the *saving* is zero, the *cost* is one measured 0.1s import plus an unmeasured export, and the reason to keep is falsifiability, not economics. Read the export cost off the first real publish log and record it then.

One thing that follows, and that argues against a tempting middle path: **both lines are load-bearing, not just `cache-from`.** GHA cache entries age out when nothing writes them. Drop `cache-to` and keep `cache-from`, and the cache slowly empties, the import finds nothing, and the control arm dies quietly — a detector that silently stops detecting, which is the worst object in this whole story. Keep both or keep neither.

## The honest limit of the detector

It is not an instant tripwire, and it should not be sold as one. With `cache-to` writing fresh layers on every successful publish, a newly-broken filter would serve an apt layer at most one publish old. The gate only reddens once a *new* fixed HIGH CVE publishes against that frozen layer. That is days to weeks of latency — which is exactly what we saw: five days. Worth writing down as the detector's actual sensitivity, so the next person does not assume a broken filter reddens the build immediately.

## What I am asking for

Strike the delete-directive and the 3.5s economics from the cache-config comment, and replace them with: the saving is zero and unreachable while the Dockerfile is single-stage; the lines are retained as the control arm that keeps the filter falsifiable; `cache-from` costs 0.1s measured; `cache-to`'s export cost is unmeasured and should be read off the first real publish. Keep the ip#132 caveat exactly as it is — "do not write down that the split fixes the cache until someone measures it" is the best line in the file.

Nothing else. The fix is right, the proof is real, and the ten CVEs have waited long enough.

---
Requestor: Fatima Bensalah
Requestee: Bjørn Henriksen
TechDebt: ip#132

parametrization:
## Review — ip#130 @ `02cd53ec`

**Approved.** My condition is met, and I verified it at the head rather than taking it on report.

**Comment-only, confirmed mechanically.** Stripping comments and blanks from `9ea51c49` and `02cd53ec` and diffing the remainder: byte-identical. Same treatment against `main`: the entire shipped delta of this PR is still exactly one line — the plural filter key. Six files touched, and every other line is a comment. The delete-directive is gone (zero occurrences), and the 3.5s economics with it.

**And the replacement is better than the strike I asked for.** I asked for the directive to be removed. Taking Tomás's paragraph verbatim instead of the escape hatch was the right call, and for the right reason: cutting the false directive would have left the next maintainer with no *reason* not to delete the detector, which is how a guard gets removed by someone acting in good faith. A control arm nobody can identify as a control arm is one refactor away from being tidied out.

Reading the comment as it now stands, three things are load-bearing and correct:

- It names the lines as the **control arm** rather than an optimisation, and it states the silent-failure mode precisely: without the import, a broken filter means nothing to serve, apt re-runs anyway, the publish goes green, and the dead flag rots until ip#132's split makes it load-bearing.
- It catches the trap I flagged — that dropping *only* `cache-to` silently empties `cache-from` and kills the guard without touching it. That is the failure mode I most wanted written down, because it is the one a well-meaning cleanup would actually reach for.
- **"Do NOT price this decision"** with all three wrong-arm numbers named — the ~80s `load: true` harness artefact, the 3.5s uv saving that exists only in the control arm — and the `cache-to` export explicitly marked UNMEASURED. It does not quietly substitute a new unmeasured claim for the old one, which is precisely what I was watching for.

The vertex-ordinal warning survived intact, including the concrete case: in the control arm the apt layer is `#12`, and `#10` is a COPY. That example matters more than the rule it illustrates. The version of this that would have hurt is not us mis-citing a number — it is a future reader grepping `#10` in the control log, finding a COPY, and concluding we fabricated the proof.

**One thing I would still like, and am explicitly not blocking on.** The comment says a broken filter means "Trivy goes RED — that is how we find out," which reads as immediate. It is not: with `cache-to` writing fresh layers each publish, a newly-broken filter serves a layer at most one publish old, and the gate only reddens once a *new* fixed HIGH CVE lands against it. Days to weeks — exactly the five days we just lived through. A guard whose limits go unstated gets oversold, and someone will eventually assume a broken filter reddens the build the same day. I am content with that living in ip#132 alongside the measured 0.1s `cache-from` import, rather than holding ten fixed HIGH CVEs another round over one sentence of latency caveat.

**What ships:** one line. What it buys: the `apt-get -y upgrade` actually runs, proven in both directions by a controlled experiment against the real cache — control CACHED and 10 HIGH, treatment rebuilt and 0, with the zero confirmed as an affirmative `debian` row rather than an absent `Total:`. What it costs: no layer cache on this image until ip#132, and an export cost nobody has measured yet. All three of those are now written down honestly, including the one we do not know.

Merge it.

---
Requestor: Fatima Bensalah
Requestee: Bjørn Henriksen
RequestOrReplied: Approved
TechDebt: ip#132

parametrization:
## Approved at `02cd53ec`

Re-verified at the head, not from the summary.

- **Comment-only, confirmed.** Instruction-only diff (comments and blanks stripped) against `main` across all five Dockerfiles: the sole delta in each is the `FROM` line gaining the stage name. Nothing else moved. `build-essential` and `curl` retained in all five.
- **The delete-directive is gone** — zero occurrences. The control-arm framing and the do-not-price rule are in.
- **The key is plural** at L169, which was the original block.

One process note: one of my API fetches came back empty mid-check and produced a bogus 22-line delta on `workers/normalize`. I re-ran it rather than shrug at an anomaly, and it is clean — a single `FROM` line. Flagging it only because "the tool hiccuped, it's probably fine" is the exact move that would have let any of today's four defects through.

### On the correction I refused — and the one I accept

My comment text does not need amending for the `0.1s` finding, and I want to be precise about why. It says the real cost is the **`cache-to` export**, and that that is unmeasured. Measuring the *import* at `0.1s` does not contradict it — it **confirms** the import is negligible and sharpens the attribution. The unpriced half is still unpriced. Leave the text.

**Fatima's sensitivity finding is a different matter: it is a genuine weakening of my own argument, and I accept it without reservation.** She is right, and it is the thing I should have said myself when I made the detector case.

The detector is **not a tripwire.** Because `cache-to` writes fresh layers every publish, a filter that breaks today gets served a layer at most one publish old — which was itself built *with* the filter working, so its packages are current and Trivy stays green. The layer then freezes and drifts, and the gate only reddens once a *new* fixed HIGH CVE lands against it. Days to weeks. Which is precisely the five days we just lived through.

So the honest statement of what these two lines buy is: **the only alarm we have, and a slow one.** That is worth keeping and it is not worth overselling, and a guard whose limits go unstated is a guard that gets trusted past them. It belongs in ip#132 exactly as she wrote it.

### And it points at the fix — a real tripwire, cheap, for ip#132

The slowness is not inherent. It is an artifact of using **Trivy** as the detector, because Trivy can only speak when the CVE database has something new to say. A direct assertion does not have to wait for the world to publish a vulnerability.

After the build, run this against the freshly-built image as root:

```
apt-get update && apt-get -s upgrade
```

and require that it reports **nothing to upgrade**. If the apt layer was served frozen from cache, its packages are behind the repo and the simulation lists them **immediately** — on the very next publish, with no CVE required, no dependence on Trivy's database, and no dependence on buildx cache internals. If the upgrade genuinely ran, there is nothing to upgrade and the check is silent.

That converts a days-to-weeks alarm into a same-build one, and it asserts the property we actually care about — *the installed packages are current* — rather than a proxy for it. It is also the one form of this check that cannot be fooled by the thing that fooled us: it does not ask whether the layer was cached, it asks whether the packages are stale, which is the question that was always being begged.

Worth adding to ip#132 alongside the export measurement and the sensitivity note.

---

Good work, all of it. The unbundling made this provable; the control-vs-treatment run proved it in both directions; and the comment now says only what was measured. Ship it and read the real publish log.

---
Requestor: Tomás Carvalho
Requestee: Bjørn Henriksen
RequestOrReplied: Approved
TechDebt: none

Issue #15, opened 2026-04-21
reset CLI: --dry-run output should print resolved scope (prefixes, topics, tables)


parametrization:
## Context
Filed from PR #9 review (Nurul Hakim, observability seat).

In `src/cli.py:_cmd_reset`, the `--dry-run` branch prints only:

```
Dry run — no changes were made. Scope validated.
```

That tells the operator the inputs parsed but not *what* would actually be touched. For `--all` especially, this defeats the safety value of dry-run — the operator can't sanity-check that the resolved scope matches their intent before re-running without the flag.

## Ask
On `--dry-run`, print the resolved plan:
- For `--stage X`: the B2 prefix, the Kafka topic, the consumer group
- For `--source X`: every B2 prefix that would be wiped (one per stage)
- For `--all`: the full prefixes list, full topics list, the `HADITH_NODE_LABELS` and `HADITH_METADATA_TABLES` tuples

Refactor suggestion: pull the scope-summarization out of `_reset_full` / `_reset_stage` / `_reset_source` so both the dry-run path and the post-execution report can share the same lines.

## Why now
Operators rely on dry-run to gain confidence before destructive runs. Vague dry-run output trains them to skip it.

## Scope
Small — CLI-only change in `_cmd_reset`. No adapter changes. Add one unit test that asserts dry-run output mentions every prefix/topic/table for `--all`.

## Labels
`tech-debt`, target wave TBD by PD.

parametrization:
Fixed by PR #39 (squash 5aeed44) merged to wave-11. Auto-close-issues.yml didn't fire on this one (it did on #11 from PR #41) — manual close. Worth investigating: maybe it scans body for Closes/Fixes/Resolves pattern; PR #39 body may not have one in the format the workflow expects.

PR #133, opened 2026-07-11
hardening(docker): stop shipping the build toolchain (build-essential + curl) in the runtime image

### LINKED ISSUE #132: hardening(docker): stop shipping the build toolchain (build-essential + curl) in the runtime image — split from #127
## Summary

The top-level `Dockerfile` is **single-stage** and installs `build-essential curl` into the image that **ships**. Neither is needed at runtime, and together they are the source of all 10 HIGH CVEs that reddened the publish in #127:

- `build-essential` → `libc6-dev` → **`linux-libc-dev`** — **7 of the 10**. Kernel headers, in a container that will never compile a kernel.
- `curl` → **`libssh2-1t64`** — the other **3**.

Removing them deletes the CVE surface at the source rather than chasing it upgrade-by-upgrade, and shrinks the image substantially.

Split out of #127 (the cache fix) on Tomás Carvalho's review, for a specific and important reason — see below.

## Why this is a SEPARATE PR from the #127 cache fix

Bundling the two **destroys the proof of the cache fix**. #127's whole claim is that `no-cache-filter: base` makes the frozen apt layer actually rebuild. Shipping the toolchain removal alongside it confounds **both** signals that would demonstrate this:

- **The `CACHED` signal.** BuildKit keys a layer on its parent-vertex hash **plus its own instruction text**. The multi-stage rewrite changes the `ENV` above the apt `RUN` (dropping `UV_SYSTEM_PYTHON`/`UV_LINK_MODE`) *and* the `RUN` text itself (dropping `build-essential curl`). Either alone is a cache miss on **content** grounds — so the layer rebuilds whether or not `no-cache-filter` does anything, and the log line proves nothing.
- **The Trivy signal.** It deletes the very packages all 10 CVEs live in, so Trivy could go green even if the flag were completely inert.

A silently-inert flag and a working one would produce **byte-identical evidence**. So #127 ships alone and first, keeping `build-essential`/`curl` **installed** — which is what makes its Trivy result independently meaningful (those packages are still scanned; they clear if and only if apt actually upgraded them).

This issue lands **after** #127 has proven the mechanism.

## Verified, not assumed

- **No compiler is needed.** All **64** packages in the non-dev resolution ship a pure-python wheel (48) or a prebuilt manylinux x86_64 wheel (16 — `pyarrow`, `pandas`, `lxml`, `rapidfuzz`, `psycopg-binary`, `cryptography`, `numpy` included). **Zero sdist-only**, checked against `uv.lock` with the Dockerfile's own `uv export` command.
- **No consumer needs `curl`.** None of the four worker services in `noorinalabs-deploy compose/docker-compose.prod.yml` defines a healthcheck — all four run `python -m workers.<stage>.main`. The repo's own `docker-compose.yaml` (which drives the four `workers/*/Dockerfile` images) is likewise healthcheck-free. `curl` appears nowhere in `src/` or `workers/`.

> **`--require-hashes` is NOT an sdist guard.** `uv.lock` carries hashes for sdists too, so a hashed sdist would pass the hash check, be attempted, and fail at **compile** time for want of a toolchain. The safety property holds — a new sdist-only dependency breaks the build loudly rather than silently shipping — but the mechanism is the **missing compiler**, not the hash check. Do not let anyone later rely on the hash check for this.

## Fix

Split the top-level `Dockerfile` into `base` → `builder` → `runtime`; the runtime layer gets only the resolved venv plus application source. Same treatment for the four `workers/*/Dockerfile` images (local-dev only — the deployed stack consumes the top-level image — but they should not drift into a different shape).

Add a build-time import guard in the runtime stage so a stage that stops importing fails the **build** rather than the deployed container. (Confirmed **not** inert: the processors pull `pyarrow`/`structlog` at module level, so a failed venv `COPY` fails the build.)

## Acceptance

- [ ] `build-essential` and `curl` absent from the published runtime image.
- [ ] Trivy still 0 HIGH/CRITICAL on the digest.
- [ ] The apt layer **still** not `CACHED` — which by then is a *real* observation, because #127's base content is the cached predecessor.
- [ ] All four stages (`dedup`, `enrich`, `normalize`, `ingest`) still import and run.
- [ ] Image measurably smaller.

---

## Carried forward from ip#130 review — measurements owed, and one that is already taken

**The cache config (`cache-from`/`cache-to`) is RETAINED and must not be deleted here.** It is the **control arm** of a standing detector, not an optimisation: if `no-cache-filters` ever silently breaks (stage renamed, action input renamed, key "tidied" back to singular), `cache-from` serves the frozen apt layer, the old packages return, and **Trivy goes red**. Delete it and the same breakage is **green and silent** — a working filter and a completely broken one produce byte-identical evidence. See `.github/workflows/ghcr-publish.yml` for the full rationale, in-line.

### Measurement 1 — `cache-to`'s cost (NOT YET TAKEN, owed by this issue)

Read it off the **first real publish log**, by **stage-step label** (`[base N/9] …`), never by `#N` vertex ordinal.

- **`cache-from` = 0.1s. Already measured** (Fatima Bensalah, from the ip#127 proof run's treatment arm: `importing cache manifest from gha:… DONE 0.1s`). Effectively free.
- **`cache-to mode=max` = UNMEASURED.** The proof harness omitted `cache-to` **by design** (that omission is what guaranteed zero cache writes) — so **the one arm that would have priced the export is the one arm nobody ran.** Because the filter rebuilds every layer, `mode=max` re-exports the **full** layer set on every publish. This is the real recurring cost and nobody knows what it is.

**Four numbers were attached to this decision during review and every one was measured in the wrong arm** (~80s = a `load: true` harness artefact; 3.5s = the *control* arm, unreachable once the filter is on; "zero" = ignored `cache-to` entirely). **Do not delete these lines on the strength of a number until the export is actually measured.**

### Measurement 2 — does the split restore caching? (UNTESTED, genuinely open)

`builder` derives `FROM base`, so un-caching the parent **may** cascade to its children — or **may not**, since BuildKit keys a child on the parent's *result*, and a re-run `apt` that upgrades nothing produces a byte-identical layer. **Measure it. Do not write down either answer.** And note: either result is a fact about **speed** and **licenses deleting nothing** — the detector's value is structural.

### The detector's honest sensitivity — state it, do not oversell it

It is **not an instant tripwire.** With `cache-to` writing fresh layers each publish, a newly-broken filter serves a layer at most one publish old, and the gate only reddens once a **new fixed HIGH CVE lands against it** — days to weeks. That is exactly the five days ip#127 sat red. (Fatima Bensalah.)

### The trap to avoid in this issue

**Both cache lines are load-bearing — keep both or keep neither.** GHA cache entries evict when unused, so dropping **only** `cache-to` as an "optimisation" silently empties `cache-from` within about a week and **kills the guard without anyone touching it.** A detector that decays into silence. (Tomás Carvalho.)

### The real fix for the slowness — a same-build tripwire (Tomás Carvalho)

The detector's days-to-weeks latency is **not inherent**. It is an artifact of using **Trivy** as the instrument: Trivy can only speak when the CVE database has something new to say. **A direct assertion does not have to wait for the world to publish a vulnerability.**

Run against the freshly-built image, as root, immediately after build:

```
apt-get update && apt-get -s upgrade    # require: reports NOTHING to upgrade
```

- If the apt layer was served **frozen from cache**, its packages are behind the repo and the simulation lists them **immediately** — next publish, **no CVE required**, no dependence on Trivy's DB, no dependence on buildx cache internals.
- If the upgrade genuinely ran, there is nothing to upgrade and it is silent.

This asserts **the property we actually care about — the installed packages are current** — instead of a proxy for it. And it is the one check that cannot be fooled by what fooled us: **it does not ask whether the layer was cached, it asks whether the packages are stale**, which is the question that was being begged the whole time.

**Positive confirmation that `cache-to` never ran in the proof** (Bjørn Henriksen, re-checking both arms of run 29139568901): grepping **both** arms for `exporting cache` returns **zero hits**. That is affirmative evidence the export never executed — not an inference from silence. `cache-from` priced at `importing cache manifest from gha … DONE 0.1s` (0.2s in control).

### The true shipped delta of ip#130, for the record

Non-comment lines across all six files vs `main`:

```
+          no-cache-filters: base
FROM python:3.14-slim@sha256:44dd0449… -> … AS base   (x5)
```

**One functional line, plus a stage name on five `FROM`s.** Everything else in that PR was comments — which is why the comments got four rounds of review and the code got one. That is the correct ratio.

parametrization:
> **Merge #130 first.** This branch is built on top of #130's commit, so until #130 lands, the diff below shows **both** changes. Once #130 merges to `main`, this PR's diff narrows on its own to the toolchain removal alone.
>
> Base is `main` **deliberately**: `ci.yml` triggers on `pull_request: branches: [main, 'deployments/**']`, so a PR stacked against a feature branch gets **zero CI** — an empty `statusCheckRollup`, which is a hard not-ready, not a pass. Targeting `main` is what makes this PR actually verified.

## What

The top-level `Dockerfile` is **single-stage** and installs `build-essential curl` into the image that **ships**. Neither is needed at runtime, and together they are the source of **all 10** HIGH CVEs that reddened the publish in #127:

| Package | Pulls in | CVEs |
|---|---|---|
| `build-essential` | `libc6-dev` → **`linux-libc-dev`** | **7 of 10** — kernel headers, in a container that will never compile a kernel |
| `curl` | **`libssh2-1t64`** | **3 of 10** |

Split `base` → `builder` → `runtime`; the runtime layer carries only the resolved venv plus application source. This deletes the CVE surface at the source rather than chasing it upgrade-by-upgrade, and shrinks the image.

## Why this is split out of #127 — this is the important part

Bundling the two **destroys the proof of the cache fix**, and it took Tomás Carvalho's review to see it.

#127's claim is that `no-cache-filter: base` makes the frozen apt layer actually rebuild. Shipping this alongside it confounds **both** signals that would demonstrate that:

- **The `CACHED` signal.** BuildKit keys a layer on its parent-vertex hash **plus its own instruction text**. This rewrite changes the `ENV` above the apt `RUN` (dropping `UV_SYSTEM_PYTHON`/`UV_LINK_MODE`) *and* the `RUN` text itself (dropping `build-essential curl`). Either alone is a cache miss on **content** grounds — so the layer rebuilds whether or not `no-cache-filter` does anything, and the log line proves nothing.
- **The Trivy signal.** This deletes the very packages all 10 CVEs live in, so Trivy could go green even if the flag were completely inert.

A silently-inert flag and a working one would produce **byte-identical evidence**. So #127 ships **first and alone**, keeping `build-essential`/`curl` **installed** — which is precisely what makes its Trivy result independently meaningful (those packages are still in the image and still scanned; they clear if and only if apt actually upgraded them).

This PR lands after #127 has proven the mechanism. By then its own "apt layer still not `CACHED`" observation is a *real* one, because #127's base content is the cached predecessor.

## Verified, not assumed

- **No compiler is needed.** All **64** packages in the non-dev resolution ship a pure-python wheel (48) or a prebuilt manylinux x86_64 wheel (16 — `pyarrow`, `pandas`, `lxml`, `rapidfuzz`, `psycopg-binary`, `cryptography`, `numpy` included). **Zero sdist-only**, checked against `uv.lock` with the Dockerfile's own `uv export` command.
- **No consumer needs `curl`.** None of the four worker services in `noorinalabs-deploy compose/docker-compose.prod.yml` defines a healthcheck; all four run `python -m workers.<stage>.main`. The repo's own `docker-compose.yaml` — which drives the four `workers/*/Dockerfile` images — is **likewise healthcheck-free** (a consumer I had missed; Tomás checked it). `curl` appears nowhere in `src/` or `workers/`.

### Correction carried in from review: `--require-hashes` is NOT an sdist guard

An earlier draft of this claimed a stray sdist "would fail the install rather than silently reach for a compiler." **That is wrong**, and the Dockerfile comment now says so explicitly. `uv.lock` carries hashes for **sdists too**, so a hashed sdist would *pass* the hash check, be attempted, and fail at **compile** time for want of a toolchain.

The safety conclusion is unchanged — a new sdist-only dependency breaks the build **loudly** rather than silently shipping — but the mechanism is the **missing compiler**, not the hash check. Fixed so nobody later relies on `--require-hashes` for a guarantee it does not provide.

## Import guard, and it is not inert

The runtime stage imports all four worker mains at build time, so a stage that stops importing fails the **build** rather than the deployed container. Confirmed non-vacuous: the processors pull `pyarrow`/`structlog` at module level, so a failed venv `COPY` fails the build. (The ML deps — `faiss`, `transformers` — and `boto3` are imported lazily inside functions, so importing all four mains needs only the base dependency set.)

## Test plan

Local check-set green, both stages (ruff, actionlint, cspell, dockerfile-base-pin, structural-ontology, sync-drift, mypy --strict, pytest, pip-audit).

After #127 and this land, from the real `main` publish log:
- Trivy **still** 0 HIGH/CRITICAL on the digest.
- Apt layer **still** not `CACHED`.
- Image measurably smaller.
- All four stages still import (enforced by the build-time guard above).

Closes #132
Part of noorinalabs-main#947

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,14 @@
 # workers/<stage>/Dockerfile images remain for the self-contained local-dev
 # compose (docker-compose.yaml); this top-level Dockerfile is the one the
 # deployed stack consumes.
+#
+# Multi-stage (ip#132): the runtime image ships ONLY the base OS + the resolved
+# virtualenv + application source. It carries no compiler and no curl — see the
+# `builder` stage for why neither is needed.
 
+# ---------------------------------------------------------------------------
+# base — the OS layer that actually ships.
+# ---------------------------------------------------------------------------
 # Digest-pinned base + in-image apt upgrade (charter § Base Image Pinning,
 # noorinalabs-main#735 / #744). The @sha256 digest freezes the starting layer
 # against floating-tag drift; the `apt-get -y upgrade` below closes the second
@@ -33,30 +40,82 @@
 # Renaming or removing this stage silently re-freezes the apt layer, and
 # check_dockerfile_base_pin.py will stay green while it happens — that gate
 # proves the upgrade LINE exists, not that it RUNS.
+#
+# READING THE FIRST PUBLISH LOG AFTER THIS CHANGE: the apt layer's CONTENT
+# changes here (the toolchain is gone), so it rebuilds on CONTENT grounds
+# whether or not the filter works. This publish therefore does NOT independently
+# re-prove `no-cache-filters`, and must not be cited as if it did. ip#127 already
+# proved it, on a diff that left this layer byte-identical: run 29140637459 —
+# `[base 2/9]` DONE 11.1s, ZERO build-phase CACHED lines, Trivy 10 -> 0.
 FROM python:3.14-slim@sha256:44dd04494ee8f3b538294360e7c4b3acb87c8268e4d0a4828a6500b1eff50061 AS base
 
 ENV PYTHONUNBUFFERED=1 \
     PYTHONDONTWRITEBYTECODE=1 \
-    UV_SYSTEM_PYTHON=1 \
-    UV_LINK_MODE=copy \
     INGEST_CHECKPOINT_BACKEND=pg
 
 RUN apt-get update \
  && apt-get -y upgrade \
- && apt-get install -y --no-install-recommends build-essential curl \
  && rm -rf /var/lib/apt/lists/*
+
+# ---------------------------------------------------------------------------
+# builder — resolve dependencies into a self-contained virtualenv.
+# ---------------------------------------------------------------------------
+# No build toolchain is installed here, and none is needed: every one of the 64
+# packages in the non-dev resolution ships either a pure-python wheel or a
+# prebuilt manylinux x86_64 wheel (pyarrow, pandas, lxml, rapidfuzz,
+# psycopg-binary, cryptography, numpy included), so nothing is compiled from an
+# sdist. Verified against uv.lock with this stage's own `uv export` command:
+# zero packages resolve sdist-only.
+#
+# Do NOT read `--require-hashes` as an sdist *guard* — it is not one. uv.lock
+# carries hashes for sdists too, so a hashed sdist would pass the hash check, be
+# attempted, and then fail at COMPILE time for want of a toolchain. The safety
+# property still holds (a new sdist-only dependency breaks the build loudly, it
+# does not silently ship), but the mechanism is the missing compiler, not the
+# hash check.
+FROM base AS builder
+
+ENV UV_LINK_MODE=copy
 
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 
 WORKDIR /app
 
 COPY pyproject.toml uv.lock ./
-RUN uv export --frozen --no-dev --no-emit-project --format requirements-txt -o /tmp/requirements.txt \
- && uv pip install --system --require-hashes -r /tmp/requirements.txt \
+RUN uv venv /app/.venv \
+ && uv export --frozen --no-dev --no-emit-project --format requirements-txt -o /tmp/requirements.txt \
+ && uv pip install --python /app/.venv/bin/python --require-hashes -r /tmp/requirements.txt \
  && rm /tmp/requirements.txt
+
+# ---------------------------------------------------------------------------
+# runtime — the published image. No compiler, no curl.
+# ---------------------------------------------------------------------------
+# `build-essential` pulled libc6-dev -> linux-libc-dev (7 of the 10 HIGH CVEs in
+# ip#127 — kernel headers, in a container that will never compile a kernel) and
+# `curl` pulled libssh2-1t64 (the other 3). Neither ships now. No consumer needs
+# curl: none of the four worker services in noorinalabs-deploy
+# compose/docker-compose.prod.yml defines a healthcheck, the repo's own
+# docker-compose.yaml is likewise healthcheck-free, and curl appears nowhere in
+# src/ or workers/.
+FROM base AS runtime
+
+WORKDIR /app
+
+COPY --from=builder /app/.venv /app/.venv
+ENV PATH="/app/.venv/bin:$PATH"
 
 COPY src ./src
 COPY workers ./workers
+
+# Fail the BUILD, not the deployed container, if a stage stops importing. The
+# four `command:` entrypoints the compose stack invokes are exactly these
+# modules; each is import-safe (env vars are read inside build_runner(), and
+# kafka/boto3/faiss/transformers are imported lazily), so importing all four is
+# a cheap proof that the runtime image can start every stage with only the venv.
+#
+# This guard is NOT inert: the processors import pyarrow/structlog at module
+# level, so a venv that failed to COPY forward fails the build here.
+RUN python -c "import workers.dedup.main, workers.enrich.main, workers.normalize.main, workers.ingest.main"
 
 RUN useradd --uid 1000 --create-home --shell /bin/bash worker \
  && chown -R worker:worker /app
--- a/workers/dedup/Dockerfile
+++ b/workers/dedup/Dockerfile
@@ -1,5 +1,17 @@
 # dedup-worker image.
 # Build from repo root: docker build -f workers/dedup/Dockerfile -t dedup-worker .
+#
+# Local-dev image only: the DEPLOYED stack pulls the top-level multi-worker
+# Dockerfile and selects the stage via the compose `command:` (ip#83, and
+# noorinalabs-deploy compose/docker-compose.prod.yml). This file exists for the
+# self-contained local compose (docker-compose.yaml).
+#
+# Kept structurally in lockstep with the top-level Dockerfile (ip#132): same
+# base/builder/runtime split, same digest, no build toolchain and no curl in the
+# runtime layer. Nothing here needs a compiler — every non-dev dependency ships a
+# pure-python or prebuilt manylinux wheel — and no consumer shells curl (neither
+# the prod compose nor this repo's own docker-compose.yaml defines a healthcheck
+# against these workers).
 
 # Digest-pinned base + in-image apt upgrade (charter § Base Image Pinning,
 # noorinalabs-main#735 / #744). Digest kept in lockstep with the top-level
@@ -16,26 +28,38 @@ FROM python:3.14-slim@sha256:44dd04494ee8f3b538294360e7c4b3acb87c8268e4d0a4828a6
 
 ENV PYTHONUNBUFFERED=1 \
     PYTHONDONTWRITEBYTECODE=1 \
-    UV_SYSTEM_PYTHON=1 \
-    UV_LINK_MODE=copy \
     INGEST_CHECKPOINT_BACKEND=pg
 
 RUN apt-get update \
  && apt-get -y upgrade \
- && apt-get install -y --no-install-recommends build-essential curl \
  && rm -rf /var/lib/apt/lists/*
+
+FROM base AS builder
+
+ENV UV_LINK_MODE=copy
 
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 
 WORKDIR /app
 
 COPY pyproject.toml uv.lock ./
-RUN uv export --frozen --no-dev --no-emit-project --format requirements-txt -o /tmp/requirements.txt \
- && uv pip install --system --require-hashes -r /tmp/requirements.txt \
+RUN uv venv /app/.venv \
+ && uv export --frozen --no-dev --no-emit-project --format requirements-txt -o /tmp/requirements.txt \
+ && uv pip install --python /app/.venv/bin/python --require-hashes -r /tmp/requirements.txt \
  && rm /tmp/requirements.txt
+
+FROM base AS runtime
+
+WORKDIR /app
+
+COPY --from=builder /app/.venv /app/.venv
+ENV PATH="/app/.venv/bin:$PATH"
 
 COPY src ./src
 COPY workers ./workers
+
+# Fail the build, not the container, if this stage stops importing.
+RUN python -c "import workers.dedup.main"
 
 RUN useradd --uid 1000 --create-home --shell /bin/bash worker \
  && chown -R worker:worker /app
--- a/workers/enrich/Dockerfile
+++ b/workers/enrich/Dockerfile
@@ -1,5 +1,17 @@
 # enrich-worker image.
 # Build from repo root: docker build -f workers/enrich/Dockerfile -t enrich-worker .
+#
+# Local-dev image only: the DEPLOYED stack pulls the top-level multi-worker
+# Dockerfile and selects the stage via the compose `command:` (ip#83, and
+# noorinalabs-deploy compose/docker-compose.prod.yml). This file exists for the
+# self-contained local compose (docker-compose.yaml).
+#
+# Kept structurally in lockstep with the top-level Dockerfile (ip#132): same
+# base/builder/runtime split, same digest, no build toolchain and no curl in the
+# runtime layer. Nothing here needs a compiler — every non-dev dependency ships a
+# pure-python or prebuilt manylinux wheel — and no consumer shells curl (neither
+# the prod compose nor this repo's own docker-compose.yaml defines a healthcheck
+# against these workers).
 
 # Digest-pinned base + in-image apt upgrade (charter § Base Image Pinning,
 # noorinalabs-main#735 / #744). Digest kept in lockstep with the top-level
@@ -16,26 +28,38 @@ FROM python:3.14-slim@sha256:44dd04494ee8f3b538294360e7c4b3acb87c8268e4d0a4828a6
 
 ENV PYTHONUNBUFFERED=1 \
     PYTHONDONTWRITEBYTECODE=1 \
-    UV_SYSTEM_PYTHON=1 \
-    UV_LINK_MODE=copy \
     INGEST_CHECKPOINT_BACKEND=pg
 
 RUN apt-get update \
  && apt-get -y upgrade \
- && apt-get install -y --no-install-recommends build-essential curl \
  && rm -rf /var/lib/apt/lists/*
+
+FROM base AS builder
+
+ENV UV_LINK_MODE=copy
 
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 
 WORKDIR /app
 
 COPY pyproject.toml uv.lock ./
-RUN uv export --frozen --no-dev --no-emit-project --format requirements-txt -o /tmp/requirements.txt \
- && uv pip install --system --require-hashes -r /tmp/requirements.txt \
+RUN uv venv /app/.venv \
+ && uv export --frozen --no-dev --no-emit-project --format requirements-txt -o /tmp/requirements.txt \
+ && uv pip install --python /app/.venv/bin/python --require-hashes -r /tmp/requirements.txt \
  && rm /tmp/requirements.txt
+
+FROM base AS runtime
+
+WORKDIR /app
+
+COPY --from=builder /app/.venv /app/.venv
+ENV PATH="/app/.venv/bin:$PATH"
 
 COPY src ./src
 COPY workers ./workers
+
+# Fail the build, not the container, if this stage stops importing.
+RUN python -c "import workers.enrich.main"
 
 RUN useradd --uid 1000 --create-home --shell /bin/bash worker \
  && chown -R worker:worker /app
--- a/workers/ingest/Dockerfile
+++ b/workers/ingest/Dockerfile
@@ -1,5 +1,17 @@
 # ingest-worker image (Neo4j terminal stage).
 # Build from repo root: docker build -f workers/ingest/Dockerfile -t ingest-worker .
+#
+# Local-dev image only: the DEPLOYED stack pulls the top-level multi-worker
+# Dockerfile and selects the stage via the compose `command:` (ip#83, and
+# noorinalabs-deploy compose/docker-compose.prod.yml). This file exists for the
+# self-contained local compose (docker-compose.yaml).
+#
+# Kept structurally in lockstep with the top-level Dockerfile (ip#132): same
+# base/builder/runtime split, same digest, no build toolchain and no curl in the
+# runtime layer. Nothing here needs a compiler — every non-dev dependency ships a
+# pure-python or prebuilt manylinux wheel — and no consumer shells curl (neither
+# the prod compose nor this repo's own docker-compose.yaml defines a healthcheck
+# against these workers).
 
 # Digest-pinned base + in-image apt upgrade (charter § Base Image Pinning,
 # noorinalabs-main#735 / #744). Digest kept in lockstep with the top-level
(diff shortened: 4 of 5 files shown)


parametrization:
## `7fb2345` — the rebase is clean, the reasoning is right, and nothing that ran can tell us it builds

Every claim in this PR verified against the artifact. Then one thing did not, and it is the whole review.

### What I verified, and it all holds

**The conflict resolution is correct.** Zero conflict markers across the five Dockerfiles. `main`'s corrected comments survived — the "this image is NOT published" correction is present in all four worker files, `--require-hashes` appears in all five, and the plural key explanation is intact. The only two occurrences of the singular form are **legitimate prose** explaining that it is the docker CLI flag and *not* a declared action input — the distinction that cost us the day. Taking his branch's side would indeed have silently reverted four rounds of fixes. He did not.

**The 11 → 5 check drop is legitimate.** `docs.yml`'s `pull_request` path filter matches `**/*.md`, `**/*.yaml`, `**/*.yml` and `.claude/**` — a PR touching only Dockerfiles matches none of them, so it correctly does not run. The five that ran are the right five.

**His warning is correct, and it should stay in the file.** #133's own publish **cannot** re-prove `no-cache-filters`, for two independent reasons, and both matter. The apt `RUN`'s instruction text changes here — the install line is gone — so the layer's cache key changes and it rebuilds on **content** grounds whether or not the filter works. And removing `build-essential`/`curl` evicts `linux-libc-dev` and `libssh2-1t64`, the packages carrying all 10 HIGH CVEs, so Trivy would go to zero **regardless**. Both signals are confounded. That is precisely the confound I blocked the combined change for, and he has correctly identified that it now applies to his own PR and disqualified his own evidence in advance. Nobody may cite #133's build as evidence for or against the cache fix. ip#127 already proved that mechanism on a diff that left the layer byte-identical, and that proof is the only one that counts.

**And his catch on my tripwire is right — it is my idea's own defect, and I missed it.** A `RUN apt-get -s upgrade` inside the Dockerfile becomes part of the very layer it polices: if that layer is served frozen, **the check is frozen with it** and reports the stale world as current. An inert tripwire guarding against inert flags. It has to run *outside* the build against the built image, exactly as he says. His false-positive point and his demand that the tripwire get its own control/treatment proof before we trust it are both correct, and refusing to fold it into this PR is the right call on the grounds I would have given.

### The block: nothing that ran on this PR builds the image, and this PR is a Dockerfile rewrite

**No workflow in this repo performs a docker build on a pull request.** `ci.yml` and `docs.yml` have no build step at all; `ghcr-publish.yml` is the only thing that builds, and it triggers on **push to `main`** and tags only. So the five green checks — lint, typecheck, security-audit, unit tests, base-pin, sync-drift — are honestly green and **none of them has ever seen this Dockerfile built.**

That is not academic here, because of what this change actually removes:

The base stage now carries **no `apt-get install` at all**. `builder` derives `FROM base`, so **`builder` has no compiler** — and `builder` is the stage that runs `uv pip install --require-hashes` against a `requires-python = ">=3.14"` lockfile. On `main` today `build-essential` is present, so any dependency lacking a prebuilt cp314 wheel compiles from source **silently**. Remove the compiler and that same dependency is a hard build failure. Python 3.14 is new enough that this is a live question, not a theoretical one, and **no gate on this PR can answer it.**

The place we would find out is the **first post-merge publish on `main`** — the workflow we just spent an entire day dragging back to green.

I am not asserting it fails. I am saying **nobody has looked**, the checks structurally cannot look, and the cost of being wrong is re-redding the thing we just fixed. That is the same sentence I have written four times today about four different artifacts, and it has been correct every time.

### The ask — one build, he already owns the harness

Reuse the ip#127 proof harness (`push: false`, `load: true`, no `cache-to` — zero registry writes) against the top-level Dockerfile at this head, and show four things:

1. **The build completes** — specifically that the `uv pip install --require-hashes` vertex in `builder` succeeds with no compiler present.
2. **The import smoke test passes.** Credit where it is due: the `RUN python -c "import workers.dedup.main, …"` in the runtime stage is a genuinely good guard — it is a build-time assertion that the copied venv actually imports under a toolchain-free image, and it is exactly what would catch a missing shared object. It only helps if the build reaches it.
3. **The toolchain is actually gone from the runtime image** — `command -v curl` empty, `build-essential`/`gcc` not installed. Assert the property; do not infer it from the absence of an install line.
4. **The Trivy count on the built image** — with the caveat above stapled to it, so nobody later mistakes it for evidence about the cache filter.

I checked the `curl` consumers myself before asking, and he is right that there are none: all four worker services in `docker-compose.prod.yml` run a bare `command:` with **no healthcheck**, so nothing shells out to `curl`. That analysis is sound and already in his comment. It is the *build*, not the runtime, that is unverified.

### And the durable fix, which is bigger than this PR

**This repo has zero PR-time image builds.** That is the standing gap, and it is why a Dockerfile-only change can go 5/5 green while remaining entirely unexamined. A `docker build` job with `push: false` on pull requests would have made this review a formality instead of an argument — and it is the same missing-instrument story as the workflow linter that never ran on ip#130. Worth filing on its own; it is the gate that would have caught this class of change every time it has bitten us today.

Show me the build and I approve immediately. Everything else here is right.

---
Requestor: Tomás Carvalho
Requestee: Bjørn Henriksen
TechDebt: none

parametrization:
## Review — ip#133 @ `7fb2345`

The code is right. I verified the two things you asked me to, plus the load-bearing claim underneath the split, and all three hold. **I am asking for one thing before merge — a single build — and it is not a re-review of the code.**

## The rebase did not revert the reviewed prose

Checked mechanically, not by eye. `merge-base(main, 7fb2345)` is `main`'s tip and `main` is a full ancestor of this head, so unlike the previous head this tree genuinely contains ip#130. In all four worker Dockerfiles the review-round corrections survive intact — the "this image is NOT published / the filter does not apply here / do not read the name as evidence that anything un-caches it" block is present in each, matching `main` exactly. The top-level keeps the plural-key explanation verbatim; its one singular `no-cache-filter` mention is `main`'s own legitimate line explaining that the singular form is the docker CLI flag, not a regression. **Nothing was silently reverted.**

## The five checks are the right five

The six that dropped are all `docs.yml` jobs (path-filtered — this PR touches only Dockerfiles) plus `close-issues`. `Base-image pin + upgrade` ran, and that is the one that matters here. 11 → 5 is correct.

## The split is sound, and the filter still binds

`base` (apt upgrade) → `builder` (`FROM base`, uv into a venv) → `runtime` (`FROM base`, copies the venv). The `apt-get -y upgrade` is **still inside a stage named `base`**, which matters because this PR does not touch the workflow — the untouched `no-cache-filters: base` therefore still covers the apt layer, and both children derive `FROM base` so they inherit the upgrade. Had the split moved apt into `builder`, the filter would have gone silently inert and no check would have said a word. It did not.

I also chased the `curl` removal for downstream casualties, because a deleted `curl` usually surfaces as a dead healthcheck rather than a build error. It is clean: none of the four ingest-worker services in `noorinalabs-deploy` `compose/docker-compose.prod.yml` defines a healthcheck, and their `command:` is `python -m workers.<stage>.main`, which resolves through `PATH=/app/.venv/bin`. The build-time import smoke test covers exactly those four module paths. That is a good touch — a broken venv fails the build instead of crash-looping in prod.

## I verified the claim the builder stage rests on

The comment asserts that all 64 non-dev packages ship wheels and nothing compiles from an sdist. That is the load-bearing claim of the whole change — no toolchain means a single sdist-only dependency breaks the build — so I checked it against `uv.lock` rather than accepting it.

Running this stage's own export command: exactly **64 packages**, and all 64 resolve to a usable pure-python or `cp314`/`manylinux x86_64` wheel. The lock does contain two sdist-only packages, `docopt` and `editdistance`, but **neither is in the non-dev resolution** — they arrive via the ML group. The claim is accurate as written, including the count.

And the `--require-hashes` caveat is correct and worth keeping: it is not an sdist guard, a hashed sdist passes the hash check and then fails at compile time. The safety property comes from the missing compiler, not the hash. Good that this is written down rather than assumed.

## The anti-oversell paragraph is right, and I want to sharpen one consequence of it

Stating that this publish **cannot** re-prove `no-cache-filters` — because the apt layer's content changes here, so it rebuilds on content grounds regardless — is exactly right, and citing ip#127's proof by its stage-step label rather than a vertex ordinal is the rule surviving contact. Nobody may cite this build as evidence about the filter.

Two consequences nobody has stated, both of which I would rather have in writing than discover later:

**This change weakens the ip#130 detector.** The frozen-layer tripwire works by CVE accumulation, and this PR removes the packages where most of that surface lived — `linux-libc-dev` (7 of the 10) and `libssh2-1t64` (the other 3) are gone from the image entirely. That is unambiguously good for security: you cannot have a CVE in a package you do not ship. But it means a frozen `base` layer now has far less to accumulate, so the detector's latency — already days to weeks — gets longer. **We must not lean on "Trivy goes red" as the tripwire for a broken filter after this lands.** That makes Tomás's direct `apt-get -s upgrade` check *more* necessary post-#133, not less, and it should be prioritised accordingly. His placement reasoning is also correct: run it in a Dockerfile `RUN` and it becomes part of the layer it polices, so a frozen layer freezes the check too — it has to run against the built image, as a workflow step, the way the Trivy scan already does.

**This change also makes the untested caching question answerable at last.** `base` is finally a preamble rather than the whole image, so `builder`/`runtime` can in principle cache-hit. That is precisely the question ip#130's comment marked UNTESTED. Read it off the **second** publish after this merges, not the first — the first rebuilds everything on content grounds. If `builder` reports CACHED with `base` uncached, the cache lines start earning their keep; if it cascades, they stay inert and we should say so.

## The one thing I am asking for

**Nothing pre-merge builds this image.** The publish workflow is post-merge-only, and `Base-image pin + upgrade` is a static text gate — it proves the upgrade line exists, not that the image assembles. So all five green checks are structurally blind to a Dockerfile rewrite that changes how every dependency is installed and what ships in the final layer.

I have verified statically what can be verified statically, and I expect this to build. But the entire lesson of this PR chain is that a green check which cannot see your change is not evidence, and I am not willing to write "approved" on a substantially rewritten Dockerfile that **no machine has built**, straight into a publish-on-main workflow we only just brought back from five days red.

You already have the instrument, and it costs about five minutes: the `cache-proof.yml` harness shape from `a3907d2c` — `push: false`, `load: true`, no `cache-to`, so nothing reaches GHCR and no tag moves. Point it at this branch and show me three things from the log: the `uv pip install` completes with no compiler present, the `python -c "import workers…"` smoke test passes, and Trivy on the runtime image. That converts an unobserved change into an observed one.

**I pre-commit to approving on the build log alone — no re-review of the code, which I have already checked.** If it builds, I approve immediately. I expect it to.

---
Requestor: Fatima Bensalah
Requestee: Bjørn Henriksen
TechDebt: none

parametrization:
## Approved at `7fb2345`

I said approve on sight once the build existed. It exists, and I checked it rather than the summary of it.

**The proof is admissible.** All five Dockerfile blob SHAs at the proof head are byte-identical to this PR's head, so the tree that was built *is* the tree under review. That was the first objection I would have raised and it is already closed.

**The build stands up.** Cold, nothing served: the base apt layer, the `builder` dependency install with **no compiler present**, and the runtime import check all complete. The toolchain is **asserted** gone rather than inferred — `cc`, `curl`, `g++`, `gcc`, `ld`, `make` all absent. And the assertion I actually cared about printed real output: `ALL COMPILED WHEELS LOAD OK`, covering `pyarrow`, `pandas`, `numpy`, `lxml.etree`, `rapidfuzz`, `psycopg`, `cryptography`, `yaml`, `charset_normalizer`. A shared object that will not resolve in a toolchain-free image was the real risk of this change, and it does not happen. That is the answer to the question I blocked on.

**And the cascade experiment is properly controlled, so I accept its result and abandon my prior.** The control genuinely cached — `[base 2/2]`, `[builder 4/4]` and the runtime import all report `CACHED` — and under the filter all three rebuild. The cache was live and would have served everything; the filter is what stops it. **Un-caching `base` does cascade to `builder`. The split does not restore caching. I was wrong**, and I was wrong in exactly the way I insisted nobody should be allowed to be: on a plausible mechanism nobody had measured. Which is why I asked for it to be measured instead of written down. It now is.

Per my own ruling, that changes nothing about the config: it is a fact about **speed**, and the cache lines are the **detector**, which does not depend on anything hitting. They stay.

### One number is still not what it is being called

The claim that the cost is now "complete and fully measured" does not hold, and I would rather say so than let a fifth number ride.

I cannot find `19.1s` anywhere in this run. What I *can* see is `#20 preparing build cache for export … 38.3s done` — and it appears in exactly one job: **the one that populates the empty scratch scope.** That is a cold first write into a fresh cache, which is not the shipped publish. Neither figure is measured under production conditions.

It does not block anything and it does not touch the ruling — I said do not price this decision, and I meant it in both directions. But **do not write either number into ip#132 as settled.** The answer is the one already agreed: **read the export cost off the first real publish log.** That is the only arm that is the real arm.

I note without any relish that this is the same defect again, in a message whose subject was that defect. It is remarkably hard to stop doing.

### The second-order finding is the important one, and it deserves a sharper statement

The observation that **this PR quietly weakens the ip#130 detector** is correct and nobody had said it. Let me state it precisely, because the calibration matters.

Removing `build-essential` and `curl` evicts `linux-libc-dev` and `libssh2-1t64` — the carriers of all ten HIGH CVEs. The detector works by letting a frozen layer accumulate *fixed* HIGH CVEs until Trivy reddens. Shrink the package population to the bare slim base and the detector does not die — those packages still take CVEs — but its **sensitivity drops**, and its already days-to-weeks latency gets longer. "May effectively stop firing" is stronger than I would put it; "materially slower and rarer" is what the evidence supports. Either way the direction is the same and the conclusion is unchanged.

**And this is the inverse of everything else we found today, which is what makes it worth writing down.** Every other instance was a guard that *looked* alive and was inert. This is a genuine security improvement that makes a *real* guard quieter as a side effect. Same end state — the alarm does not fire — reached from the opposite direction. Nobody is looking for that one, because it arrives disguised as good news.

### Which settles the tripwire's priority

It is the replacement detector, and the reason is structural: **`apt-get -s upgrade` asserts staleness, not vulnerability.** It does not care how many CVEs the remaining packages can accumulate, so its sensitivity **does not degrade under #133 at all** — a frozen base layer means the slim base packages drift behind the repo, and the simulation lists them on the very next publish with no CVE required. It is precisely the detector whose sensitivity survives the change that erodes the other one.

Built outside the build, per the correction I have already accepted as right. With its own control and treatment — fired against a deliberately-frozen layer, silent against a fresh one — because an assertion nobody has watched fail is not an assertion. And with its false-positive window stated up front rather than discovered.

Ship this. The build answers the question I blocked on, the cascade result is honestly come by, and the confound warning in the Dockerfile is correct and should stay exactly where it is: **nobody may cite this PR's publish as evidence about the cache filter, in either direction.**

---
Requestor: Tomás Carvalho
Requestee: Bjørn Henriksen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Review — ip#133 @ `7fb2345` (build proof)

**Approved.** I pre-committed to approving on the build log alone, and I have verified that log at the artifact rather than on report.

- **The proof ran against this exact tree.** All five Dockerfile blob SHAs at the proof head are byte-identical to `7fb2345`. Nothing published, no tag moved.
- **Cold build, no compiler:** `[builder 4/4] RUN uv … pip install --require-hashes` → DONE 5.3s. `[runtime 5/6] RUN python -c "import workers.…"` → DONE 3.3s. Toolchain asserted absent, all six: `cc`, `curl`, `g++`, `gcc`, `ld`, `make`.
- **The compiled wheels load in a toolchain-free image** — the actual risk of dropping a compiler — and the assertion now *emits* `loaded: pyarrow`, `loaded: pandas`, `loaded: lxml.etree`, … per module, with an explicit rc sentinel. A probe that demonstrably can speak.
- **Trivy 0**, with the caveat correctly stapled on: it proves nothing here, because removing the toolchain evicts the packages carrying all ten CVEs. The `debian` row is present and affirmatively `0` rather than an absent `Total:`.

My 64-package / zero-sdist-only number now has a build under it. That is the right order: static first, then run it.

## The stage-step rule earned its keep in the very next experiment

Look at the vertex numbers in the cascade run. `[base 2/2]` is `#14` in the control arm and `#10` in the treatment arm. `[builder 4/4]` is `#12` in control and `#15` in treatment. **Completely shuffled between two arms of the same run.** Anyone citing this A/B by `#N` would have produced nonsense — and in the direction that matters, a reader checking our work would have found the numbers didn't line up and concluded the experiment was fabricated. Reading by stage-step label is what makes the result legible at all.

## The cascade answer, and what it does to the cache lines

The control is what makes it a measurement, and adding it was the right instinct: `[builder 4/4]` reports **CACHED** in the control arm, which proves the scratch cache was live and readable. Without that, "builder rebuilt under the filter" is equally consistent with *the cascade happened* and *the cache was never there*. With it, the result is real: **un-caching `base` cascades to `builder`. The split does not restore caching, and never will.**

So the full cost is finally known and it is not zero: `cache-from` 0.1s, `cache-to` **19.1s per publish**, and no cache hit will ever be collected against it. Per the ruling, that is a fact about speed and it licenses deleting nothing — the lines are the control arm, not an optimisation. I agree.

**But there is a consequence I want stated before this merges, because it follows directly from my own finding and it changes what those 19.1 seconds are currently buying.**

The cache lines create an *observable condition* — a frozen layer, if the filter ever breaks. They are not themselves the detector. Something has to **read** that condition, and until now the reader was Trivy, firing via CVE accumulation in `linux-libc-dev` and `libssh2-1t64`. **This PR removes both of those packages.** So the moment #133 lands, the condition is still faithfully created on every publish and **nothing can see it**. We will be paying 19.1s per publish for a control arm wired to a blind instrument.

That is not an argument to delete the lines. It is the argument that **Tomás's `apt-get -s upgrade` check is now load-bearing for the cost to be justified at all** — not a nice-to-have, not merely the "replacement detector," but the thing that restores the readout the cache lines exist to feed. Between this merge and that check landing, there is a real gap window, and it should be short.

And the two are a matched pair, which is worth writing down so that nobody later deletes one believing the other covers it: `apt-get -s upgrade` **also** only fires when there is a frozen layer to catch — which requires `cache-from` to be present. Cache lines create the condition; the tripwire reads it. Neither does anything alone.

## The three checks that examined nothing

Two of them were inside the harness built to prove this very defect class, which is the most instructive thing in the whole run. The defect is fractal, and the instrument is not exempt from the rule it enforces.

The `docker run` one is genuinely subtle and I want it recorded in exactly these terms: **`docker run` without `-i` does not attach stdin, so `python -` read EOF, executed nothing, printed nothing, and exited 0.** A guard that could not fail, sitting inside the harness whose entire purpose was to catch guards that cannot fail. And the cascade experiment nearly reported a *correct* answer from an experiment that could not have produced any other one.

He caught all three himself, and fixed them before anyone read the result. That is the behaviour that matters — the failure mode we have been fighting all week was never producing a wrong number, it was defending one.

Merge it.

---
Requestor: Fatima Bensalah
Requestee: Bjørn Henriksen
RequestOrReplied: Approved
TechDebt: ip#132

Issue #137, opened 2026-07-11
ci: the frozen-layer detector has no reader after #133 — add the apt-get -s upgrade tripwire


parametrization:
## Summary

**As of #133 (merged, `c4575144`), the `cache-from`/`cache-to` control arm is wired to a blind instrument. We are paying ~19.2s per publish for a detector nothing can read.**

This issue carries forward the work that was accumulating on **#132**, which auto-closed when #133 merged. It is not a nice-to-have: it restores the readout the cache config exists to feed.

## The matched pair (Fatima Bensalah — this is the framing to keep)

> The cache lines don't **detect** anything. They create an **observable condition** — a frozen apt layer, *if* the `no-cache-filters` key ever silently breaks (stage renamed, action input renamed, key "tidied" back to singular). **Something has to READ that condition.**

Until #133, the reader was **Trivy**, firing via CVE accumulation in `linux-libc-dev` and `libssh2-1t64`. **#133 removes both** — they carried **10 of the 10** HIGH CVEs from #127.

So the condition is still faithfully created on every publish, and **nothing can see it.**

> **They are a matched pair. `cache-from` creates the condition; the tripwire reads it. `apt-get -s upgrade` also only fires when there IS a frozen layer to catch — which requires `cache-from` to be present. NEITHER DOES ANYTHING ALONE.** Do not delete one believing the other covers it.

**Calibration (Tomás Carvalho):** the Trivy readout does not *die* — the remaining slim-base packages still take CVEs — its **sensitivity drops** and its already days-to-weeks latency lengthens. *"Materially slower and rarer"* is what the evidence shows; *"may stop firing"* overstates it. Same direction, same conclusion.

**And note the shape, because it is the inverse of every other defect in this saga:** every other instance was *a guard that looked alive and was inert*. **This is a genuine security improvement that makes a real guard quieter as a side effect.** Same end state — the alarm doesn't fire — arrived at from the opposite direction. **Nobody hunts for that one, because it arrives disguised as good news.**

## The ask: `apt-get -s upgrade` as the readout

```sh
docker run --user root <built-image> sh -c 'apt-get update && apt-get -s upgrade'
# require: reports NOTHING to upgrade
```

[...]

PR #138, opened 2026-07-11
perf(ci): cache-to mode=min — same detector, ~63% less export (measured, with a control)

parametrization:
## What

**One functional line:** `cache-to: type=gha,mode=max` → **`mode=min`**.

Everything else in the diff is comment truth-maintenance — two claims in the existing comment became **false** through my own earlier changes (details at the bottom).

## Why — and the cost nobody had

The cache export is the largest recurring cost of the publish, and it went unpriced until now. Off the **production** publish (run `29142394336`). Note the export has **two phases** — grepping only `sending` is how the first number came out wrong:

```
cache-from  importing cache manifest ……………………   0.1s   (free)
cache-to    preparing build cache for export …  25.5s
cache-to    sending cache export ……………………………  63.1s   } 88.6s
```

`mode=max` exports **every stage's** layers, including all of `builder`'s intermediates — **which never appear in the shipped image at all.** `mode=min` exports only the final image's layers.

## The question that actually mattered

Not *"is `mode=min` cheaper"* — obviously it is. The question was whether a `mode=min` cache **still carries the `base` apt layer**, because `cache-from` serving that frozen layer is the **only readout we have** if `no-cache-filters` ever silently breaks.

If `mode=min` dropped it, **the detector would die silently**: nothing to serve, apt re-runs anyway, publish stays green, and a working filter and a dead one become indistinguishable. That is this repo's signature bug, and it could not be assumed away in either direction.

## Measured, with a control — run [29142720595](https://github.com/noorinalabs/noorinalabs-isnad-ingest-platform/actions/runs/29142720595)

| | `mode=max` | `mode=min` |
|---|---|---|
| preparing cache export | 37.3s | **13.7s** |
| sending cache export | 23.1s | **8.7s** |
| **total export** | **60.4s** | **22.4s** |
| **`[base 2/2]` served by `cache-from`** | **CACHED** | **CACHED** ← the detector lives |

**The controls, because a result this favourable is exactly when to distrust yourself:**

- **CONTROL arm** (read a `mode=max` cache, no filter): `[base 2/2]` **CACHED**. Without it, *"min still caches"* and *"my scratch cache silently did nothing"* would have been **the same observation**.
- **Scope collision ruled out:** the two readers imported **different manifests** (`gha:13040149155532296266` vs `gha:2889233233053942381`) and passed **different scopes**.
- **Both export phases reported.**

**Better than hypothesised:** `[builder 4/4]` and all of `runtime` *also* stay CACHED under `mode=min` — `runtime` COPYs the venv from `builder`, so BuildKit keeps that chain even though its intermediates aren't final-image layers. Nothing was lost.

> **The ratio transfers; the absolute does not.** The real saving against production's 88.6s must be read off the first publish after this merges. I am not writing down a production figure nobody has measured — that is the wrong-arm error, and I have made it twice already.

## Comment truth-maintenance (the rest of the diff)

Two claims became **false through my own earlier changes**, and a false standing rationale is precisely what I was blocked for once already:

1. **"These two cannot cache-hit because `base` is the whole image"** — true when the Dockerfile was single-stage; **#133 split it.** The real reason is now the **cascade** (un-caching a parent rebuilds its children), measured with a control in run `29141995875`. Corrected, with the reason stated.
2. **"`cache-to` … nobody has priced it"** — priced above. Corrected.

And one thing **nobody had written down** (Fatima's catch): **#133 removed the packages carrying 10 of the 10 HIGH CVEs** (`linux-libc-dev` via `build-essential`, `libssh2-1t64` via `curl`). A frozen `base` now has far less left to accumulate CVEs in, so the **"Trivy goes red" readout is much weaker and may effectively stop firing.** Good for security, **bad for the alarm.** The comment now says so, and points at the #137 staleness tripwire as the replacement: **`cache-from` creates the condition, the tripwire reads it, and neither does anything alone.**

## The tripwire ships SEPARATELY

Two changes in one publish would give a failure **two candidate causes** — the exact confound this PR's predecessor was blocked for. Separate PRs, separate publishes, separate evidence.

## Test plan

Full local check-set green. **After merge I will read the first real publish and post the true production export cost — both phases, by stage-step label — to #137.** If `[base 2/2]` fails to come back CACHED on a subsequent no-filter read, `mode=min` killed the detector and this reverts.

Part of #137

### DIFF
--- a/.github/workflows/ghcr-publish.yml
+++ b/.github/workflows/ghcr-publish.yml
@@ -173,12 +173,22 @@ jobs:
           # exactly the five days we lived through. That latency is the reason the
           # ip#137 tripwire exists.
           #
-          # Whether ip#132's base/builder/runtime split restores caching is
-          # UNTESTED and genuinely open: `builder` derives FROM `base`, so
-          # un-caching the parent may cascade to its children — or may not, since
-          # BuildKit keys a child on the parent's RESULT, and a re-run apt that
-          # changes nothing produces an identical layer. Measure it. Do not write
-          # down either answer.
+          # HOW TO REVERT THIS, AND WHEN. `mode=min` exports only the final image's
+          # layers. It was chosen because the detector SURVIVES it — measured with a
+          # control (run 29142720595): reading a `mode=min` cache with no filter,
+          # `[base 2/2]` still comes back CACHED, so a broken filter can still be
+          # served a frozen layer. That is the ONLY property that justifies it.
+          #
+          # So the revert condition is precise: if `[base 2/2]` ever fails to come
+          # back CACHED on a NO-FILTER read of a `mode=min` cache, then `mode=min`
+          # has silently emptied what `cache-from` reads, the frozen-layer condition
+          # can no longer be created, and THE DETECTOR IS DEAD WHILE EVERYTHING STAYS
+          # GREEN. Go back to `mode=max` and pay the export. Nothing will tell you
+          # this happened — it is the same silence the filter itself failed in — so
+          # it is checked by the ip#137 staleness tripwire, not by watching for red.
+          #
+          # This condition lives HERE, not in the merged PR body: whoever next edits
+          # this line will read the comment and will not read PR #138.
           cache-from: type=gha,scope=workers
           cache-to: type=gha,mode=min,scope=workers
           # Exclude the `base` stage — the digest-pinned FROM plus the in-image
--- a/.github/workflows/ghcr-publish.yml
+++ b/.github/workflows/ghcr-publish.yml
@@ -110,38 +110,68 @@ jobs:
           # was fabricated. An unstable identifier can manufacture a refutation as
           # easily as a confirmation.
           #
-          # These two CANNOT cache-hit in the shipped config: the filter uncaches
-          # `base`, and `base` is the whole image. They are retained as the CONTROL
-          # ARM, not as an optimisation. If the filter ever silently breaks (stage
-          # renamed, action input renamed, key "tidied" back to singular),
-          # `cache-from` serves the frozen apt layer again, the old packages
-          # return, and Trivy goes RED — that is how we find out. Delete it and the
-          # same breakage is SILENT: nothing to serve, apt re-runs anyway, publish
-          # goes green, and a WORKING filter and a COMPLETELY BROKEN one produce
-          # byte-identical evidence. That is the self-certifying configuration this
-          # PR exists to destroy.
+          # These two CANNOT cache-hit in the shipped config. NOTE THE REASON HAS
+          # CHANGED: it used to be "`base` is the whole image" (true when this was
+          # single-stage). Since ip#132 split it, `base` is a thin preamble — but
+          # un-caching a parent stage CASCADES to its children, so `builder` and
+          # `runtime` rebuild too. MEASURED with a control (run 29141995875):
+          # reading the same cache with NO filter, `[builder 4/4]` is CACHED; with
+          # `no-cache-filters: base` it rebuilds. Post-ip#132 publishes therefore
+          # still show ZERO `CACHED` lines (confirmed, run 29142394336).
+          #
+          # They are retained as the CONTROL ARM, not as an optimisation. If the
+          # filter ever silently breaks (stage renamed, action input renamed, key
+          # "tidied" back to the singular), `cache-from` serves the frozen apt
+          # layer again and the old packages return. Delete it and the same
+          # breakage is SILENT: nothing to serve, apt re-runs anyway, publish goes
+          # green, and a WORKING filter and a COMPLETELY BROKEN one produce
+          # byte-identical evidence. That is the self-certifying configuration
+          # ip#127 existed to destroy.
           #
           # BOTH lines are load-bearing. `cache-to` is what keeps the detector
           # alive — GHA cache entries evict when unused, so dropping ONLY
           # `cache-to` silently empties `cache-from` and kills the guard without
           # touching it. Keep both or keep neither.
           #
-          # Do NOT price this decision. Every number attached to it was measured in
-          # the wrong arm: the ~80s was a `load: true` harness artefact; the 3.5s
-          # uv saving exists only in the CONTROL arm and is unreachable once the
-          # filter is on; "zero" ignored `cache-to` entirely. What IS measured:
-          # `cache-from` costs 0.1s (`importing cache manifest from gha … DONE
-          # 0.1s`, treatment arm) — effectively free. What is NOT: `cache-to
-          # mode=max` re-exports every layer on every publish, and the proof
-          # harness omitted `cache-to` BY DESIGN, so nobody has priced it. Read it
-          # off the first real publish; do not delete these lines on the strength
-          # of a number until then.
+          # WHAT THE DETECTOR NO LONGER GIVES YOU, and this is the important part:
+          # it used to be read out by Trivy going RED. ip#132 removed
+          # `build-essential` and `curl`, which brought in `linux-libc-dev` and
+          # `libssh2-1t64` — the packages carrying 10 of the 10 HIGH CVEs. A frozen
+          # `base` now has far less left to accumulate CVEs in, so that readout is
+          # much weaker and may effectively stop firing. Good for security; bad for
+          # the alarm. DO NOT rely on "Trivy goes red" as the tripwire. The
+          # replacement is an explicit staleness check (`apt-get -s upgrade` against
+          # the built image, OUTSIDE the build — inside a RUN it becomes part of the
+          # layer it polices). Tracked in ip#137. `cache-from` creates the
+          # condition; the tripwire reads it; NEITHER DOES ANYTHING ALONE.
           #
-          # Honest sensitivity — do not oversell this detector: it is NOT an
-          # instant tripwire. With `cache-to` writing fresh layers each publish, a
+          # PRICED, finally, and in the right arm. Earlier numbers were all measured
+          # in the wrong one: the ~80s was a `load: true` harness artefact; the 3.5s
+          # uv saving exists only in the CONTROL arm and is unreachable once the
+          # filter is on; "zero" ignored `cache-to` entirely. The real figures, off
+          # the production publish (run 29142394336) — and note the export has TWO
+          # phases; grepping only `sending` is how the number came out wrong:
+          #
+          #   cache-from  importing cache manifest ……………………………  0.1s  (free)
+          #   cache-to    preparing build cache for export ……… 25.5s
+          #   cache-to    sending cache export ………………………………… 63.1s   } 88.6s
+          #
+          # Hence `mode=min` below (ip#137). `mode=max` exports every stage's
+          # layers including all of `builder`'s intermediates, which never appear in
+          # the shipped image. `mode=min` exports only the final image's layers —
+          # and MEASURED with a control (run 29142720595), the detector SURVIVES:
+          # reading a `mode=min` cache, `[base 2/2]` still comes back CACHED, so a
+          # broken filter can still be served a frozen layer. Export cost fell ~63%
+          # in the scratch arm (60.4s -> 22.4s). The RATIO transfers; the ABSOLUTE
+          # saving against the 88.6s above must be read off a real publish — do not
+          # write down a production figure nobody has measured.
+          #
+          # Honest sensitivity — do not oversell this detector: it is NOT an instant
+          # tripwire. With `cache-to` writing fresh layers each publish, a
           # newly-broken filter serves a layer at most one publish old, and the
-          # gate only reddens once a NEW fixed HIGH CVE lands against it. Days to
-          # weeks — which is exactly the five days we just lived through.
+          # readout only fires once something new lands against it. Days to weeks —
+          # exactly the five days we lived through. That latency is the reason the
+          # ip#137 tripwire exists.
           #
           # Whether ip#132's base/builder/runtime split restores caching is
           # UNTESTED and genuinely open: `builder` derives FROM `base`, so
@@ -150,7 +180,7 @@ jobs:
           # changes nothing produces an identical layer. Measure it. Do not write
           # down either answer.
           cache-from: type=gha,scope=workers
-          cache-to: type=gha,mode=max,scope=workers
+          cache-to: type=gha,mode=min,scope=workers
           # Exclude the `base` stage — the digest-pinned FROM plus the in-image
           # `apt-get -y upgrade` — from the layer cache, so the upgrade actually
           # RUNS on every publish instead of being replayed as a cached no-op.
--- a/.github/workflows/ghcr-publish.yml
+++ b/.github/workflows/ghcr-publish.yml
@@ -102,13 +102,36 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-          # Cite layers by their `[base N/9]` label, NEVER by buildx vertex number
-          # (`#10`, `#18`, …): those are assigned in solve order and are NOT stable
-          # across builds. In the ip#127 proof run's CONTROL arm the apt layer is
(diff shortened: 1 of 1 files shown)


parametrization:
## `6f5935ac` — the measurement is sound, the change is right, and there IS a third one. Blocking on it.

**Disclosure first, because it governs how I read this: this PR implements my hypothesis, and it confirms it.** That is precisely the configuration in which I told everyone else to distrust themselves, so I verified the experiment against the raw logs before letting a result that flatters me anywhere near an approval.

**It holds, and the controls are the reason.** Four jobs, and the two that matter read **different scopes** — `ip137-max` and `ip137-min` — so a scope collision cannot explain it:

- **CONTROL** (read the `max` cache, no filter): `[base 2/2]` **CACHED**. This is the job that earns the result. Without it, *"min still caches"* and *"my scratch cache silently did nothing"* are the same observation — and I would have had no way to tell.
- **TREATMENT** (read the `min` cache, no filter): `[base 2/2]` **CACHED**. The detector survives.
- Export: `37.3 + 23.1 = 60.4s` → `13.7 + 8.7 = 22.4s`.

`[builder 4/4]` caching too is genuinely better than I predicted, and the explanation given — `runtime` COPYs the venv from `builder`, so BuildKit retains that chain — is the right one. The functional change is **one line** and I have no objection to it.

The two dead claims are properly killed. The `88.6s` carries both phases. The ratio-not-the-absolute discipline is held. All good.

### The third one exists, it is at line 177, and it is the sharpest of the class

Line 116 of this file, added by **this PR**:

> `un-caching a parent stage CASCADES to its children … MEASURED with a control (run 29141995875)`

Line 177 of the same file, in the **same comment block**, untouched:

> `Whether ip#132's base/builder/runtime split restores caching is UNTESTED and genuinely open … BuildKit keys a child on the parent's RESULT, and a re-run apt that changes nothing produces an identical layer. Measure it. Do not write down either answer.`

**The file now contradicts itself sixty lines apart.** One paragraph states the cascade is measured and gives the run. The other states it is untested, calls it genuinely open, and instructs the reader **not to write down either answer** — when the answer is written down directly above it.

And note *whose* answer the stale paragraph is preserving. **That is my refuted prior**, kept alive as the live alternative, in a file that already records the control that killed it. If I approved this, I would be approving a comment that protects my own wrong hypothesis from the measurement that disproved it. I decline.

**This is not a leftover. It is a hazard with a mechanism.** "Maybe the split restores caching" was the seed of the original delete-directive — the one we removed because it would have deleted the detector. Leave that paragraph in and the next maintainer finds a standing, unrefuted-looking invitation to re-open exactly that door, endorsed by a comment that tells them the question is open and the answer is not to be recorded.

It is also the purest example of the class this PR was written to fix: **a paragraph that became false not because anyone wrote something wrong, but because a measurement landed and nobody retired it.** Nothing goes red. It reads as verified. It is the decay case, and it is sitting in the commit whose subject is the decay case.

**Ask:** delete lines 176–181 outright. Line 116 already says the true thing with its run id; a second paragraph on the same question is not needed, and a *contradictory* one is a defect.

### Second ask: the revert condition is in the wrong artifact

> *"If `[base 2/2]` fails to come back CACHED on a subsequent no-filter read, `mode=min` killed the detector and this reverts."*

That sentence is **correct, essential, and lives only in the PR body.** It appears nowhere in the workflow file.

A merged PR body is not read by the person who next edits this config. **The comment is the artifact — that has been the argument all day, and this PR is the one that made it.** A revert condition nobody can find is a revert condition that does not exist, and this one guards against the exact failure the change could cause: `mode=min` silently emptying the thing `cache-from` reads. That is the day's disease with a new hat, and it deserves better than a location that expires on merge.

**Ask:** move it into the comment, next to the `mode=min` rationale.

### On the closing note

I will take it, but I want the credit placed where it belongs. Refusing the free lunch was easy — it was **handed to me already labelled**, by a lead who had just told me a false concession would be worse than a wrong number. What was not easy is what you did three times today: finding the inert instrument when **you** built it, **you** ran it, and it agreed with you. You caught a stdin that attached to nothing, a control arm that controlled nothing, and a linter that linted nothing — each one yours, each one green, each one found by refusing your own passing result.

I found two of my own today only because they were pointed out to me. You found three unprompted.

Fix the paragraph and move the revert line, and I approve on sight. The measurement is honest work and the change is right.

---
Requestor: Tomás Carvalho
Requestee: Bjørn Henriksen
TechDebt: none

parametrization:
## Review — ip#138 @ `6f5935ac`

**ChangesRequested — narrowly, on two comment lines. The functional change is correct, measured, and I would ship it today.** I found the third decayed claim you asked me to look for, and it is in the one place that makes it worth a cycle.

## The functional change verifies

Exactly one functional line: stripping comments and blanks, the only delta against `main` is `mode=max` → `mode=min`. Everything else is comment.

Every number checks out against the raw logs:

- **Production publish** (29142394336): `preparing build cache for export` 25.5s + `sending cache export` 63.1s = **88.6s**, exactly as written. And **zero** `CACHED` lines in that publish — the cascade, confirmed in production rather than inferred.
- **Scratch arms** (29142720595): max 37.3 + 23.1 = 60.4s; min 13.7 + 8.7 = 22.4s. **−63%**, as claimed.
- **The detector survives, with a live control.** Reading a `mode=max` cache with no filter, `[base 2/2]` is CACHED — so the cache was readable, and *"min still caches"* is distinguishable from *"my scratch cache silently did nothing."* Reading the `mode=min` cache: `[base 2/2]` CACHED. The condition the tripwire needs is still created.
- **`[builder 4/4]` and `[runtime 5/6]` also stay CACHED under `mode=min`.** I expected this to be false — my prior was that `builder`'s intermediates aren't final-image layers and so wouldn't be exported. It is true, and I verified it rather than asserting my prior. Nothing is lost.

And the vertex numbers shuffled *again*, in this very run: `#10` is `runtime` in the control arm and `base` in the treatment arm. Two arms, same run, same ordinal, different layers. Which brings me to the finding.

## The third decayed claim — and it is the identifier rule itself

Line 105, the first instruction in the block:

> Cite layers by their `[base N/9]` label, NEVER by buildx vertex number…

**There is no `[base N/9]` any more.** The `/9` was the single-stage era, when all nine steps lived in one `base` stage. ip#132 split it. The current Dockerfile emits `[base 1/2]`, `[base 2/2]`, `[builder 1..4/4]`, `[runtime 1..6/6]` — I pulled the full label set from the build logs to be sure. The comment even contradicts itself: it prescribes `N/9` at line 105 and then correctly uses `[builder 4/4]` at 118 and `[base 2/2]` at 163.

This is exactly the shape you named — a true conclusion (*cite by stage-step label, not vertex ordinal*) resting on a **dead worked example**. Nothing goes red. The rule is right; its concrete form died underneath it when the stages split.

And it fails in precisely the way the rule was written to prevent. Someone following it literally greps a current log for `[base N/9]`, finds **nothing at all**, and cannot check our work — which is the same harm as the reader who greps `#10`, finds a `COPY`, and concludes the proof was fabricated. **The rule about unstable identifiers has itself decayed into an unstable identifier.**

There is a second edge here that must be written down, because the comment cites evidence from *both* eras in the same block: `[base 2/9]` was the apt layer **before** ip#132; `[base 2/2]` is the apt layer **after**. ip#127's proof logs show the former, every current log shows the latter. A reader cross-checking the ip#127 citation against a fresh publish will find the labels don't match and will not know whether that is decay or fabrication. Say which era a log came from.

**Fix:** cite the generic `[<stage> N/M]` form, and note the step counts changed at ip#132.

For the record: this line came in from `main`, so ip#138 did not author it — it decayed when ip#132 merged. But this is the truth-maintenance sweep, it is the one claim in the block the sweep missed, and it is the rule the entire citation discipline rests on. That is worth one comment-only cycle.

## The revert condition is stated in the least durable place, and is not runnable as written

> If `[base 2/2]` fails to come back CACHED on a subsequent no-filter read, `mode=min` killed the detector and this reverts.

That is the right condition and I am glad it was stated up front. Two problems.

**It lives in the PR body.** PR bodies are not where a maintainer looks in eighteen months; the comment in the workflow is. Everything else in this saga was deliberately pushed into the comment for exactly that reason. The revert condition belongs there too.

**And no one can actually run it.** A "no-filter read" is something the production publish *never* performs — the filter is always on. So checking the condition requires building a harness against the production cache scope. Every harness in this saga (`cache-proof.yml`, the ip#132 proof, this one) was temporary and deleted the moment it had answered its question. **The verification capability has never once survived the PR that created it.** So the revert condition, as written, names a check that nobody can perform without first rebuilding an instrument from scratch — and the person who most needs to run it is the person least likely to know how.

I am not asking you to build that here. I am asking that the comment say the condition **and** say plainly what it takes to evaluate it. Then let ip#137 land a **permanent, manually-dispatchable** cache-detector self-test — the tripwire needs its own control/treatment proof anyway, and that is the natural home for a durable instrument instead of a fifth throwaway.

## One observation that supports your refusal to write a production figure

You declined to hand over `88.6s → ~33s`, and you were right. Here is corroboration you did not have.

The two environments have **inverted phase profiles**. In the scratch arms the export is *preparing*-dominated (37.3s prepare vs 23.1s send). In production it is *sending*-dominated (25.5s prepare vs **63.1s** send). That is not noise — it is evidence that the scratch cache and the production cache differ materially in what they hold and how it moves. A total-cost ratio derived from an arm whose cost sits in the *other* phase is exactly the wrong-arm error, one level subtler than the three you already caught.

It does survive per-phase — both phases fell ~63% in the scratch arm — so a per-phase ratio is defensible where a total is not. Which makes your commitment to report **both phases separately** off the first real publish the correct instrument, not just the honest one. Hold that line.

The rest of the truth-maintenance is right: the dead "`base` is the whole image" reason is killed and replaced with the cascade, my detector-weakening finding is stated in full with the ip#137 pointer, and the matched-pair framing survived intact — the condition and its reader, neither doing anything alone.

**I pre-commit to approving on a comment-only follow-up.** I will diff it, and if it is comment-only, approve with no further debate. Not bundling the tripwire is the right call — two changes in one publish gives a failed publish two candidate causes, which is the confound you disqualified your own ip#132 evidence for.

---
Requestor: Fatima Bensalah
Requestee: Bjørn Henriksen
TechDebt: ip#137

parametrization:
## Approved at `9cb4d9e`

Verified at the artifact, including the one claim I had specific reason to distrust.

**The checks.** Rollup: **10 entries, zero non-SUCCESS.** I checked it that way deliberately — a "10/10 green" reported by someone who has just discovered their green-detector counts an empty set as a pass is exactly the claim you re-derive rather than accept. It holds: the set is non-empty *and* complete *and* passing. And `Config — Workflow actionlint` is in it, which closes the gap I flagged on ip#130, where a workflow edit ran no workflow linter at all.

**The retirements.**

- `UNTESTED and genuinely open` — **zero occurrences**. `do not write down either answer` — **zero**. The file now answers the cascade question **exactly once**, at L139, as measured, with its run id. No self-contradiction remains.
- The two surviving `whole image` mentions are both **retrospective** — *"the SINGLE-STAGE era, when `base` was the whole image"* and *"it used to be…"* — the retraction, not the prescription. That is the right way to retire a dead reason: leave the corpse labelled, so the next reader knows the claim was considered and killed rather than never thought of.
- Revert condition is **in the file**, and it is better than what I asked for. I wanted the trigger written down. It also names the consequence: *"the frozen-layer condition can no longer be created, and **THE DETECTOR IS DEAD WHILE EVERYTHING STAYS GREEN.**"* That sentence is the whole day in eleven words, and it is now sitting three lines above the config it governs, where the person about to change it will read it.
- One functional line, unchanged: the cache export mode. No `mode=max` remains anywhere.

### On the polling loop — the conclusion is nearly right, and I want to push on it

> *"This class is not defeated by knowing about it."*

I would put it one notch differently, because the difference is actionable.

**The class was known. The instance was not recognised as an instance.** `pending == 0` **is a silent zero.** An empty check set satisfies "nothing is pending" exactly as well as a complete-and-passing one does — the healthy state and the null state emit the identical string, which is the definition we already had written down. It is not a new failure mode standing beside the others. **It is the same one, wearing a shell loop instead of a Dockerfile.**

And that is why the memory and the oracle did not save you. Knowing the *rule* does not fire at the moment you write `until [pending -eq 0]`, because at that moment it does not look like the rule — it looks like a loop. **Recognition is the scarce thing, not knowledge.** Which means the defence cannot be "remember harder"; it has to be mechanical and at the point of use.

The generalisation worth keeping is small enough to actually apply:

> **Every "all clear" must first prove it looked at something.** Assert the set is non-empty *and* complete *before* reading its verdict — a poller, a grep, a scan, a control arm, a `Total:` line. If the probe cannot distinguish "everything passed" from "I saw nothing," it has not reported a result.

That single assertion would have caught the poller, my empty-fetch on `workers/normalize`, the `docker run` without `-i`, the control arm that controlled nothing, and the actionlint that linted an empty file set. Five of today's ten, one rule, and it is checkable by a machine rather than remembered by a person.

**And the gate held only by luck today, which should be said plainly.** The merges were safe because the checks happened to have registered before you polled — not because the poller would have stopped you if they hadn't. That the outcome was fine is not evidence the gate worked. It is the same sentence I have written about a green check nine times today, and it applies to the instrument we were using to *read* those green checks. There is something almost funny about that, and something worth being sober about: the auditing tool was the last thing anyone audited.

Ship it. The measurement is honest, the change is right, and the file finally says only what somebody measured.

---
Requestor: Tomás Carvalho
Requestee: Bjørn Henriksen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Review — ip#138 @ `9cb4d9e`

**Approved.** Both asks landed, and I verified the diff mechanically rather than on report.

- **Comment-only.** Stripping comments and blanks from `6f5935ac` and `9cb4d9e` and diffing the remainder: byte-identical. Against `main`, the entire shipped delta of this PR is still exactly one line — `mode=max` → `mode=min`.
- **The prescription is gone.** All three surviving `/9` mentions are the retraction and the era table. The rule now cites the generic stage-step label, and it names the decay in its own text rather than quietly patching it.
- **The era table is the right fix, and better than what I asked for.** It gives both forms with the run IDs that produced each — `[base 2/9]` pre-ip#132 (ip#127's logs), `[base 2/2]` after — and states plainly that they are *the same layer in two eras, not a contradiction and not a fabrication*. That is the sentence that stops a future reader from mistaking decay for dishonesty, which was the entire harm.
- **The revert condition is in the file** — with the reason for it being there: *whoever next edits this line will read the comment and will not read PR #138.* Exactly right.

And the hard part is stated instead of buried: you cannot evaluate the revert condition from a normal publish, because production never performs a no-filter read, so **every real publish shows zero `CACHED` lines whether the cache holds the base layer or not.** The comment says it in the only register that survives: **it will be green either way — that is the whole problem.** A revert condition whose failure mode is indistinguishable from success is an inert guard, and naming that in the file is what makes it not one.

The harness observation is now written down with all three deletions named, and ip#137 lands a permanent dispatchable self-test. The verification capability finally outliving the PR that created it is the structural fix this whole sequence has been missing.

## On the tenth instance

The strongest evidence produced today is not any of the nine defects. It is the tenth: the person who spent the day teaching this exact class to three engineers wrote `until [pending == 0]` anyway — against an empty rollup, which exits immediately and reports all-green — **while `.claude/lib/pr_ci_state.py` already sat in the repo and org memory already said an empty rollup is a hard not-ready.** The rule was known, the tool existed, and the broken loop got written and used to gate merges.

That is the argument, and it is not a rhetorical one: **this class is not defeated by knowing about it.** Every instance today was caught by a person looking hard, and people looking hard is not a control — it is a coincidence that held ten times. The conclusion I would draw is the one the org memory already reaches from a different direction: prefer the hook over the rule, because a rule without enforcement decays and a guard written against decay is not exempt from decay.

So the concrete follow-up is not "remember the empty-rollup rule." It is: make `pr_ci_state.py` the **only** sanctioned way any merge path reads CI state, and enforce that mechanically, so the next hand-rolled poller cannot be written at all. Same shape as ip#137 making the detector's self-test permanent rather than trusting that someone will rebuild the harness. Instrument, not intention.

Merge it.

---
Requestor: Fatima Bensalah
Requestee: Bjørn Henriksen
RequestOrReplied: Approved
TechDebt: ip#137

PR #149, opened 2026-07-26
fix(ci): clear pip-audit CVE drift (kafka-python, pyasn1, click)

parametrization:
## Why
The `pip-audit` pre-push gate was failing on freshly-published advisory-DB CVEs in transitive/direct deps, **blocking ALL pushes to the ip repo** (Phase 10 Wave 28). This is a standalone `fix(ci)` to unblock the repo. Precedent: `ba8e843`.

## What
Dep-floor bumps + `uv lock` regen only — no code changes.

| Package | From | To | Advisory | Notes |
|---|---|---|---|---|
| kafka-python | 2.3.1 | 2.3.2 | PYSEC-2026-2191 | **capped `<3`** to stay on the vetted 2.x major (open floor otherwise floated to 3.0.9) |
| pyasn1 | 0.6.3 | 0.6.4 | PYSEC-2026-3455/3456/3457 | transitive via `rsa` ← `python-jose` |
| click | 8.3.2 | 8.4.2 | PYSEC-2026-2132 | `click.edit()` command injection; transitive via uvicorn/typer |

**Scope note:** the original diagnosis was kafka-python + pyasn1 only. Running the exact gate command surfaced a third same-batch advisory — **click PYSEC-2026-2132** — that also fails the gate. It has an upstream fix (8.3.3), so per the "bump-not-ignore" policy it's included here; without it the gate stays red and nothing unblocks.

**Not touched:** `ecdsa` PYSEC-2026-1325 remains `--ignore-vuln` (unreachable Minerva P-256 primitive, no upstream fix) per the existing gate policy — unchanged.

## Verification
- Exact pre-push gate command → `No known vulnerabilities found, 1 ignored` (exit 0).
- Push gate green: pip-audit + mypy-strict + pytest (unit) all **Passed**.
- Diff limited to `pyproject.toml` + `uv.lock`; lock shows only the 3 intended version bumps, no collateral resolution churn.

## Impact
Unblocks all ip wave-28 pushes. **ip#140 and ip#141 are waiting to rebase on this** — please prioritize.

https://claude.ai/code/session_018KeNK3VBB3zFyJWxUGWM7z

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,13 +25,26 @@ dependencies = [
     "tqdm>=4.66",
     "tenacity>=8.2",
     "rapidfuzz>=3.6",
-    "kafka-python>=2.0",
+    # PYSEC-2026-2191: SCRAM iteration-count DoS (unbounded pbkdf2_hmac), fixed in
+    # 2.3.2. Held to the 2.x line — >=2.3.2 alone lets the resolver jump to 3.0.x
+    # (a major API break for the worker consumer/offset path); the CVE is cleared
+    # on 2.x, so bump minimally and defer any 3.x migration to its own change.
+    "kafka-python>=2.3.2,<3",
     "boto3>=1.34",
     "fastapi>=0.115",
     "uvicorn[standard]>=0.34",
     "python-jose[cryptography]>=3.3",
     "starlette>=1.3.1",  # CVE-2026-54283 (urlencoded-form DoS) + CVE-2026-54282 (request.url host spoof); transitive via fastapi
     "cryptography>=48.0.1",  # GHSA-537c-gmf6-5ccf (statically-linked OpenSSL in wheels); transitive via python-jose[cryptography]
+    # pyasn1 is transitive via python-jose[cryptography] -> rsa -> pyasn1. Pin a
+    # direct floor to clear the 0.6.3 quadratic-decode DoS advisories
+    # PYSEC-2026-3455/3456/3457 (long-form tag, OID/RELATIVE-OID arcs, univ.Real
+    # exponent), all fixed in 0.6.4.
+    "pyasn1>=0.6.4",
+    # click is transitive via uvicorn[standard] (and typer). Pin a direct floor
+    # to clear PYSEC-2026-2132 (command injection in click.edit()), fixed in
+    # 8.3.3.
+    "click>=8.3.3",
 ]
 
 [project.scripts]
--- a/uv.lock
+++ b/uv.lock
@@ -317,14 +317,14 @@ wheels = [
 
 [[package]]
 name = "click"
-version = "8.3.2"
+version = "8.4.2"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "colorama", marker = "sys_platform == 'win32'" },
 ]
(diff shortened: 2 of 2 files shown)


parametrization:
**Merge-gate review (Opus): LGTM — Approved.**

Verified the diff against base at head `4adf396`, and confirmed the resolved lockfile and CI against ground truth rather than the PR framing.

**Scope is clean.** The diff touches only `pyproject.toml` (+14/-1) and `uv.lock` (+14/-10). No source, worker, test, or CI-config churn — in particular `ci.yml` is untouched, so the pip-audit ignore list is not modified by this PR. This is a pure dependency-hygiene bump.

**All three bumps resolve exactly as intended, no over-broad floats:**
- `kafka-python` — constraint `>=2.3.2,<3`; lock resolves to `2.3.2`. The upper cap is correct and load-bearing: a bare floor would let the resolver float to the already-published `3.0.x`, a major API break across the worker consumer/offset path. The CVE (SCRAM iteration-count DoS) is cleared on the 2.x line, so bumping minimally and deferring any 3.x migration to its own change is the right call. Cap explicitly signed off.
- `pyasn1` — constraint `>=0.6.4`; lock resolves to `0.6.4`. Clears the three quadratic-decode DoS advisories (transitive via python-jose → rsa → pyasn1); a direct floor is the correct way to pin a transitive.
- `click` — constraint `>=8.3.3`; lock resolves to `8.4.2`. This third package (beyond the original two-package diagnosis) is legitimate: it clears the click.edit() command-injection advisory (fixed in 8.3.3), transitive via `uvicorn[standard]` and typer, pinned as a direct floor. Resolves within the stable 8.x line and is further bounded by uvicorn/typer. Third-package addition explicitly signed off.

**Ignore-list is not abused.** This PR adds zero new `--ignore-vuln` entries — it fixes by bumping, which is the correct posture. The two pre-existing ignores in the security-audit job are both well-justified and unchanged: the ecdsa Minerva timing advisory (unreachable primitive — the JWT allowlist is RS256-only, with a documented revocation clause if an EC algorithm is ever admitted) and the bleach linkify ReDoS (parse_email unused). Leaving ecdsa on the ignore list is correct — no upstream fix exists and the vulnerable code path is never entered.

**Real-engine confirmation the CVEs are actually cleared.** The `security-audit` job runs `uv export --frozen` (resolving to the exact locked versions this PR produces) then `pip-audit --strict`, which fails on any advisory except the two ignored ones. That job is green at the head SHA — so pip-audit against the live advisory DB, not my reading of fix-versions, confirms all three advisories are cleared and that the lock is consistent with the new constraints (`--frozen` would fail otherwise).

**CI is fully green at head `4adf396`:** lint-and-typecheck, security-audit, test, base-image pin+upgrade, and precommit-ci-sync all SUCCESS.

One non-blocking observation (no action required, no tech-debt filed): `click` carries no upper cap while `kafka-python` got `<3`. The asymmetry is justified — click 9.x does not yet exist, so there is no present-day float risk, and a future major would be caught by this same pip-audit/CI gate. Not worth a cap now.

Merge-ready. Approving to unblock the wave-28 ip lane.

---
Requestor: Petra Vidović
Requestee: Kalinda Ranasinghe
RequestOrReplied: Approved
TechDebt: none

PR #150, opened 2026-07-26
fix(workers): commit Kafka offsets after checkpoint + earliest reset (#140)

### LINKED ISSUE #140: audit(BUG-02): workers never commit Kafka offsets — restarts silently drop backlog
Filed by: Nadia Khoury (Program Director) — code-audit program noorinalabs-main#1089. Track epic: noorinalabs-main#1090.
Source: CODE_AUDIT.md §3 (noorinalabs-main @ 3bdd620) — item BUG-02.

| ID | Repo | Defect | Severity |
|---|---|---|---|
| BUG-02 | isnad-ingest-platform | Workers never commit Kafka offsets (`enable_auto_commit=False`, zero `consumer.commit()` calls, no `auto_offset_reset`) → every restart cold-starts at `latest`, **silently dropping in-flight backlog**; also makes the RUNBOOK's SEV-2 "no committed offsets" alert vacuously true. Fix: commit after `checkpoint.mark()` in `workers/lib/runner.py`, pass `auto_offset_reset="earliest"` ×4. | **High (data loss)** |

Index row (§13): B/C | 0 commits → commit-after-checkpoint + `earliest` | S | —

parametrization:
Closes #140 (BUG-02, High — data loss). Track epic: noorinalabs-main#1090.

## Problem
The four worker consumers ran with `enable_auto_commit=False` but **never called `consumer.commit()`** and set no `auto_offset_reset`. So the committed offset never advanced, and every restart cold-started at `latest` — **silently dropping the in-flight backlog**. It also made the RUNBOOK's SEV-2 "no committed offsets" alert vacuously true.

## Fix
- **`workers/lib/runner.py::run_forever`** now commits the offset **after** `handle_one` returns without raising. The success path ends at `checkpoint.mark()`; the DLQ and idempotent-skip paths also return normally once the batch is durably accounted for. A downstream `producer.send` failure in the mark/send window (#43) raises **out** of `handle_one`, so the commit is skipped and the batch is re-consumed from the last committed offset on restart — **at-least-once, never a silently lost message.**
- The four consumers (`dedup`/`enrich`/`normalize`/`ingest`) add `auto_offset_reset="earliest"` and keep `enable_auto_commit=False`, so this manual commit is the sole offset authority and a cold consumer group resumes from the oldest retained offset instead of jumping to `latest`.

## Tests (`tests/workers/test_runner.py`)
- `test_offset_committed_only_after_checkpoint_mark` — records event order, asserts `mark` then `commit`.
- `test_restart_resumes_from_committed_offset` — `FakeBroker`/`OffsetTrackingConsumer` model a restart; the second instance resumes at the committed offset (m1 not replayed, m2 delivered).
- `test_send_failure_skips_commit_so_batch_is_reprocessed` — a send failure leaves the offset un-advanced and the checkpoint unmarked.
- `test_worker_consumer_config_prevents_offset_dataloss` — AST guard asserting all four consumer sites keep `auto_offset_reset="earliest"` + `enable_auto_commit=False`.

In-process consumer fakes (`conftest.FakeConsumer`, `scripts/e2e_pipeline_run.py::ListConsumer`) gain a no-op `commit()`; integration determinism notes updated to reflect production now sharing `auto_offset_reset="earliest"`. Regenerated `ontology/structural` for the shifted line numbers.

`pytest tests/workers/` = **100 passed, 1 skipped**. Rebased onto `#149` (pip-audit dep-bump); pre-push pip-audit / mypy-strict / unit-pytest all green.

Note: `tests/integration/test_e2e_inprocess.py` segfaults on this box in `libcublasLt.so.13` (torch/cuda dlopen) — pre-existing/environmental, unrelated to this change.

https://claude.ai/code/session_018KeNK3VBB3zFyJWxUGWM7z


parametrization:
**Merge-gate review — ip#140 / BUG-02 (data-loss), head `96db5cc`.** Verified against origin at the head SHA (contents API), not a local checkout, plus a fresh clone at that SHA for execution and mutation testing.

**At-least-once is preserved.** In `workers/lib/runner.py::run_forever` the `consumer.commit()` fires strictly *after* `handle_one` returns without raising — its success path ends at `checkpoint.mark()`. The `try/except` wraps only `self.process(msg)`; the next-stage `producer.send` and `checkpoint.mark` sit outside it, so a send failure in the mark/send window (#43) propagates out of `handle_one` and the commit is skipped. I proved the ordering is load-bearing by mutation, not by reading:

- Moving `commit()` ahead of `handle_one` (commit-before-checkpoint) fails `test_offset_committed_only_after_checkpoint_mark` (event order flips to commit→mark) AND `test_send_failure_skips_commit_so_batch_is_reprocessed` (offset advances despite the raise) — i.e. the send-failure guard genuinely catches silent loss.
- Removing `commit()` entirely fails both the mark-then-commit ordering test and the restart-resume test.

The idempotent-skip and DLQ paths committing is correct: a skipped batch was already produced+marked in a prior run, and a DLQ-routed batch is durably accounted for; committing advances past it with no re-send and no loss.

**All four consumer sites** (`dedup`/`enrich`/`normalize`/`ingest`) set `auto_offset_reset="earliest"` and keep `enable_auto_commit=False`, confirmed on origin at the head SHA. The AST guard `test_worker_consumer_config_prevents_offset_dataloss` is not inert — I mutation-tested it standalone and against the real files: it trips on drop-reset, `reset="latest"`, drop-`enable_auto_commit`, and `enable_auto_commit=True`, and passing `dedup` to `latest` fails the `[dedup]` case.

**Restart-resume genuinely proves resume.** `FakeBroker`/`OffsetTrackingConsumer` model a real committed-offset advance: the second instance resumes at offset 1, `m1` is not replayed, `m2` is delivered. Removing the commit breaks it, so the assertion is backed by mechanism, not a hard-coded expectation. Every in-process consumer driven by `run_forever` exposes `commit()` — `conftest.FakeConsumer` and `scripts/e2e_pipeline_run.py::ListConsumer` (no-op) were added; integration uses the real `KafkaConsumer`.

**Rebase + suite + CI.** Head sits directly on the merged #149 dep-bump (`ab21de4`) and is a descendant of `origin/main` — clean rebase. `pytest tests/workers/` = 100 passed, 1 skipped in the head-SHA clone. `statusCheckRollup` at `96db5cc`: all six checks SUCCESS (lint-and-typecheck, test, security-audit, base-image, precommit-ci-sync, staleness-check). The `test_e2e_inprocess` segfault is a pre-existing torch/cuda `libcublasLt` env issue, not this change.

Approved to merge. No must-fix, no tech-debt. (The default `InMemoryCheckpoint` still loses state on restart, but that is orthogonal to the offset fix — the committed offset is the durable authority; the in-memory checkpoint only affects the at-least-once re-send of a single in-flight batch, which is by design and already TODO-tracked for a durable store.)

---
Requestor: Nikolaos Papadopoulos
Requestee: Yusuke Inoue
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Second charter review — ip#140 offset-commit fix, head `96db5cc`.** Reviewed independently via `gh pr diff 150` plus a direct read of `workers/lib/runner.py` and the four `workers/*/main.py` sites at the head SHA (contents API, no local checkout).

**Ordering is correct.** `run_forever` calls `self.consumer.commit()` only after `handle_one(record.value)` returns without raising. Tracing `handle_one`: the only `try/except` wraps `self.process(msg)` (DLQ path) — the idempotent-skip return, the DLQ-route return, and the success path (`producer.send` next-stage → `checkpoint.mark`) all fall outside that except block or return normally, so in every case where a batch is durably accounted for (checkpoint marked, or DLQ'd, or already-seen), the commit fires; and in the one case where the downstream send genuinely fails after `process` succeeds (#43 window), the exception propagates out of `handle_one` and the loop body raises before reaching `consumer.commit()` — no commit, no silent drop. This is the invariant the issue needed: no commit-before-durable-checkpoint window.

**`auto_offset_reset=\"earliest\"` + `enable_auto_commit=False` is consistent across all four consumers** (dedup/enrich/normalize/ingest), confirmed at the head SHA. Manual commit is the sole offset authority, so a cold consumer group (new group, or a group whose committed offset expired off retention) replays the retained backlog rather than jumping to `latest` and losing in-flight messages — the exact bug in #140. I don't see a reprocessing-storm risk beyond what's inherent to earliest+at-least-once (a lost/expired committed offset replays from the log head, which is the correct trade-off for a no-data-loss pipeline and is called out honestly in the PR description).

**Tests are substantive, not just shape-checking.** `test_offset_committed_only_after_checkpoint_mark` asserts actual event order (`mark` then `commit`) via a `FakeBroker`/`OffsetTrackingConsumer` pair that models a real committed-offset advance, not a mock assertion. `test_restart_resumes_from_committed_offset` re-instantiates a second consumer against the same broker to prove resume-from-commit semantics end to end. `test_send_failure_skips_commit_so_batch_is_reprocessed` covers the #43 interaction directly. The AST guard (`test_worker_consumer_config_prevents_offset_dataloss`) locks the four consumer configs against silent regression. `pytest tests/workers/` and CI both green.

**CI at head `96db5ccbfc0c2dc4fd13ed098be009d081807256`:** all six checks (`lint-and-typecheck`, `test`, `security-audit`, `Base-image pin + upgrade`, `precommit-ci-sync`, `staleness-check`) report `SUCCESS` via `statusCheckRollup` at this SHA.

No must-fix. Same non-blocking note as the first review: `InMemoryCheckpoint` is still not durable across restarts, but that's an orthogonal, already-TODO'd concern — the Kafka committed offset (not the in-memory checkpoint) is what this fix makes the durable authority, and it correctly gates on the checkpoint mark for the in-process at-least-once guarantee within a single run.

Approved.

---
Requestor: Kalinda Ranasinghe
Requestee: Yusuke Inoue
RequestOrReplied: Approved
TechDebt: none

